PR #23, opened 2026-03-08
Internal event bus

davesnowdon:
Not accessible externally yet. Just getting ready for forthcoming features which will needs events

### DIFF
--- a/src/lithos/server.py
+++ b/src/lithos/server.py
@@ -77,6 +77,13 @@ class LithosServer:
         """Get configuration."""
         return self._config
 
+    async def _emit(self, event: LithosEvent) -> None:
+        """Emit an event, logging any failure without propagating."""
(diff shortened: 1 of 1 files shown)


src/lithos/server.py
```diff
@@ -155,15 +169,35 @@ async def handle_file_change(self, path: Path, deleted: bool = False) -> None:
...
+                            try:
+                                await self.event_bus.emit(
+                                    LithosEvent(
+                                        type=NOTE_DELETED,
+                                        payload={"path": str(relative_path)},
+                                    )
+                                )
+                            except Exception:
+                                logger.exception("Failed to emit event for file watcher delete")
                     else:
```
davesnowdon: lot of repeated boilerplate that could be in a helper method

Issue #32, opened 2026-03-11
[CRITICAL] Pickle deserialization - arbitrary code execution risk


davesnowdon:
## Summary
`pickle.load()` on untrusted file enables arbitrary code execution in graph cache loading.

## Audit Reference
- **Audit ID:** 1.1
- **Severity:** 🔴 Critical
- **Category:** Security

## Affected Files
- `src/lithos/graph.py` - `load_cache()` function

## Problem
The graph cache uses `pickle.load()` to deserialize the NetworkX graph from disk. In Docker deployments with bind mounts, any process with write access to the data directory can plant a malicious `graph.pickle` file. When Lithos loads this file, it will execute arbitrary Python code.

This is a well-known attack vector - pickle deserialization is explicitly unsafe for untrusted data.

## Recommendation
Replace pickle serialization with JSON using NetworkX's built-in functions:
- `nx.node_link_data()` for serialization
- `nx.node_link_graph()` for deserialization

This provides the same functionality without the arbitrary code execution risk.

## Additional Context
- OWASP: Deserialization of Untrusted Data (A8:2017)
- Python docs explicitly warn against `pickle.load()` on untrusted data

PR #52, opened 2026-03-11
fix: replace pickle serialization with JSON in graph cache

hanumanclaw:
## Summary

Eliminates an arbitrary code execution risk by replacing unsafe `pickle` serialization in the knowledge graph cache with `json`.

## Changes

- `src/lithos/graph.py`: Replace `pickle` import with `json`. Change `graph_cache_path` to return `graph.json` instead of `graph.pickle`. Use `json.load`/`json.dump` with `nx.node_link_graph`/`nx.node_link_data` for NetworkX graph (de)serialization. Keep atomic write pattern (tempfile + os.replace).
- `tests/test_integration_conformance.py`: Update cache filename reference from `graph.pickle` to `graph.json`.

## Security

Pickle deserialization is a well-known arbitrary code execution vector — any attacker who can write to the cache file can execute arbitrary code. JSON has no such risk.

Fixes agent-lore/lithos#32

### DIFF
--- a/src/lithos/graph.py
+++ b/src/lithos/graph.py
@@ -94,9 +94,10 @@ class KnowledgeGraph:
                 )
                 return False
             graph_data = data.get("graph")
-            self._graph = (
-                nx.node_link_graph(graph_data, edges="links") if graph_data else nx.DiGraph()
-            )
+            if graph_data and "nodes" in graph_data and "links" in graph_data:
+                self._graph = nx.node_link_graph(graph_data, edges="links")
+            else:
+                self._graph = nx.DiGraph()
             self._id_to_node = data.get("id_to_node", {})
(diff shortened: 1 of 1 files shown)


hanumanclaw (commented):

- Replaces pickle serialisation with JSON in the graph cache ( → )
- Adds `GRAPH_CACHE_VERSION = 1` guard with a warning-and-rebuild path on version mismatch
- Uses `nx.node_link_graph(..., edges="links")` which requires networkx >=3.2 — correctly reflected in the pin update in `pyproject.toml`
- Round-trip serialisation tests added
- Old `graph.pickle` files are silently ignored on first run (new cache path is different); no migration needed

**No issues found.** The version guard correctly absorbs what was previously tracked in #61. The networkx constraint is tight enough to guarantee the `edges=` kwarg is available.

Issue #36, opened 2026-03-11
[HIGH] Write operations are not atomic across storage layers


davesnowdon:
## Summary
Multi-step write operations can leave indexes in inconsistent state on crash or error.

## Audit Reference
- **Audit ID:** 2.2
- **Severity:** 🔴 High
- **Category:** Architecture

## Affected Files
- `src/lithos/knowledge.py`
- `src/lithos/search.py`
- `src/lithos/graph.py`

## Problem
`lithos_write` involves multiple sequential operations:
1. Write file to disk
2. Update internal maps
3. Release lock
4. Index in Tantivy
5. Index in ChromaDB
6. Update knowledge graph

Steps after lock release are not transactional. If the process crashes mid-operation, the system is left in an inconsistent state:
- File exists but not indexed
- Tantivy indexed but not ChromaDB
- Indexes updated but graph stale

On restart, there's no automatic detection or repair of this drift.

## Recommendation
1. Implement auto-reconcile on startup if dirty state detected
2. Consider write-ahead log (WAL) for pending operations
3. Add periodic consistency check between file system and indexes
4. The existing `lithos reconcile` command could be enhanced to run automatically on startup

## Additional Context
This is a fundamental challenge for multi-store systems. A pragmatic approach is "eventual consistency with startup reconciliation" rather than full distributed transactions.

PR #54, opened 2026-03-11
fix: make knowledge document writes atomic using write-then-rename

hanumanclaw:
## Summary

Implements atomic write semantics for knowledge document persistence. A crash mid-write no longer leaves corrupt or partial files on disk.

## Changes

- `src/lithos/knowledge.py`: Add `_atomic_write(path, content)` helper that writes to a temp file in the same directory then renames it into place (POSIX rename is atomic on same filesystem). Replace the two `full_path.write_text()` calls in `create()` and `update()` with `_atomic_write()`. On failure the temp file is cleaned up.

## Pattern

The graph cache (`graph.py`) already used this pattern. This PR applies the same approach to the markdown knowledge store.

Write → temp file → `os.replace(tmp, dest)` → atomically visible or not at all.

Fixes agent-lore/lithos#36

### DIFF
--- a/tests/test_knowledge.py
+++ b/tests/test_knowledge.py
@@ -12,12 +12,44 @@ from lithos.knowledge import (
     KnowledgeManager,
     KnowledgeMetadata,
     WriteResult,
+    _atomic_write,
     generate_slug,
     normalize_derived_from_ids_lenient,
     normalize_url,
     parse_wiki_links,
     validate_derived_from_ids,
 )
+
+
+class TestAtomicWrite:
+    """Tests for _atomic_write helper."""
+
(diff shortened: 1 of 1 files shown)


hanumanclaw:

hanumanclaw (commented):

- Replaces direct `write()` calls with an atomic write-then-rename pattern (write to temp file, `os.replace` into final path)
- Prevents partial writes from leaving corrupt documents on disk in the event of a crash or signal mid-write
- Cleanup test added to verify temp files are not left behind on success or failure

**No issues found.** The atomic rename approach is correct on POSIX and Windows (Python 3.3+). The temp file is created in the same directory as the target, ensuring the rename is same-filesystem.

Issue #45, opened 2026-03-11
[HIGH] No optimistic locking on lithos_write


davesnowdon:
## Summary
Concurrent updates to the same document result in silent last-writer-wins with no conflict detection.

## Audit Reference
- **Audit ID:** 4.1
- **Severity:** 🔴 High
- **Category:** Product/Usability

## Affected Files
- `src/lithos/server.py` - `lithos_write` tool
- `src/lithos/knowledge.py`

## Problem
When two agents update the same document concurrently:
1. Agent A reads document (updated_at: T1)
2. Agent B reads document (updated_at: T1)
3. Agent A writes changes (updated_at: T2)
4. Agent B writes changes (updated_at: T3) - **silently overwrites A's changes**

There's no way for agents to detect this conflict. Agent A's work is lost without any indication.

## Recommendation
Add optional optimistic locking:
1. New parameter `if_match_updated_at: datetime | None` on `lithos_write`
2. If provided and document's `updated_at` doesn't match, return error:
   ```json
   {"error": "stale_write_conflict", "current_updated_at": "..."}
   ```
3. Agent can then re-read, merge changes, and retry
4. When not provided, current last-writer-wins behavior preserved (backwards compatible)

## Additional Context
This is standard HTTP ETag/If-Match semantics adapted for the MCP context. Essential for multi-agent collaboration on shared documents.

PR #55, opened 2026-03-11
fix: add optimistic locking via version field in frontmatter

hanumanclaw:
## Summary

Implements optimistic locking on `lithos_write` to prevent silent concurrent overwrites. Writers can now supply the version they read, and the server will reject the write if a newer version has been committed since.

## Changes

- `src/lithos/knowledge.py`:
  - Add `version: int = 1` field to `KnowledgeMetadata` and `_KNOWN_METADATA_KEYS`.
  - `from_dict` reads `version` (defaults to 1 for existing docs without the field).
  - `_parse_version` clamps to `max(1, int(value))` so non-positive frontmatter values fall back to 1.
  - `to_dict` serializes `version` to frontmatter.
  - `update()` gains optional `expected_version: int | None` parameter.
  - If `expected_version` is provided and doesn't match the on-disk version, returns `WriteResult(status="error", error_code="version_conflict")` without writing.
  - On a successful write, increments `version` by 1 (bumped just before `_atomic_write` so early returns leave the version unchanged).
- `src/lithos/server.py`: Expose `expected_version` as an optional parameter on the `lithos_write` MCP tool.
- `tests/test_knowledge.py`: New test class `TestOptimisticLocking` with full coverage of the knowledge-layer behaviour.
- `tests/test_server.py`: New test class `TestOptimisticLockingServerLayer` with full coverage of the MCP server layer.

## Behaviour

| Caller passes `expected_version` | On-disk version | Result |
|---|---|---|
| `None` (default) | any | Write succeeds (backwards compatible) |
| matches | matches | Write succeeds, version incremented |
| stale | newer | `error / version_conflict` returned |

Fixes agent-lore/lithos#45

### DIFF
--- a/tests/test_knowledge.py
+++ b/tests/test_knowledge.py
@@ -3100,6 +3100,7 @@ class TestOptimisticLocking:
         assert result.message is not None
         assert "1" in result.message
         assert "2" in result.message
+        assert result.current_version == 2
 
     @pytest.mark.asyncio
     async def test_correct_expected_version_succeeds(self, knowledge_manager: KnowledgeManager):
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1242,3 +1242,16 @@ class TestOptimisticLockingServerLayer:
             agent="agent",
         )
         assert update_result["status"] == "updated"
+
+    @pytest.mark.asyncio
+    async def test_expected_version_ignored_on_create(self, server: LithosServer):
+        """expected_version is silently ignored on create (not an error)."""
+        result = await self._call_write(
+            server,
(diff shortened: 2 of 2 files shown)


hanumanclaw (commented):

- Adds a `version` field to the knowledge document frontmatter schema for optimistic locking
- Callers pass an expected version; a mismatch raises a conflict error before any write occurs
- CHANGELOG entry added documenting the frontmatter schema change (breaking: new `version` field added to frontmatter)

**No issues found.** The version increment-and-check logic is correct. The CHANGELOG entry makes the schema change discoverable for downstream consumers.

Issue #37, opened 2026-03-11
[MEDIUM] FastMCP omit-vs-null ambiguity is structurally unsound


davesnowdon:
## Summary
The `_UNSET` sentinel pattern exists to work around FastMCP's inability to distinguish omitted fields from null values.

## Audit Reference
- **Audit ID:** 2.3
- **Severity:** 🟡 Medium
- **Category:** Architecture

## Affected Files
- `src/lithos/server.py`
- `src/lithos/knowledge.py`

## Problem
FastMCP delivers both "field omitted" and "field set to null" as Python `None`. This creates ambiguity:
- Agent omits `tags` field → wants to leave tags unchanged
- Agent sets `tags: null` → wants to clear tags

The current `_UNSET` sentinel pattern works but creates maintenance hazards:
- Easy to forget sentinel handling in new tools
- Inconsistent patterns across different tools
- Not self-documenting in type signatures

## Recommendation
1. Add `MaybeUnset[T]` type alias to formalize the pattern
2. Create helper functions for consistent handling
3. Document the pattern clearly for contributors
4. File upstream issue with FastMCP to request proper omit/null distinction

## Additional Context
This is a common issue with JSON-based protocols. GraphQL solves it with explicit `null` vs field absence. FastMCP may need similar semantics.

PR #58, opened 2026-03-11
fix: formalise _UNSET sentinel pattern for all optional update fields

hanumanclaw:
## Summary

Resolves the structural ambiguity where FastMCP conflates omitted fields (`None`) with explicitly null values, making it impossible to distinguish "not provided" from "explicitly null".

## Problem

FastMCP maps both omitted parameters and explicit JSON `null` to Python `None`. Without a sentinel, `lithos_write`'s update path couldn't distinguish "preserve existing tags" from "clear all tags".

Previously, `tags` and `confidence` passed `None` directly to `KnowledgeManager.update()`, where `None` meant "clear" — so omitting `tags` on an update would silently wipe them.

## Changes

- `src/lithos/knowledge.py`: Changed `update()` signature for `tags` and `confidence` to use `_UnsetType` default (`= _UNSET`) instead of `None`. Added docstring entries explaining sentinel semantics for both fields.
- `src/lithos/server.py`: In the update path of `lithos_write`, map `tags=None` → `_UNSET` (preserve) and `confidence=None` → `_UNSET` (preserve). Updated docstring to document sentinel conventions for all optional update fields.
- Tests: Added coverage verifying that omitting `tags`/`confidence` on update preserves existing values, and that passing `[]` / a new float sets/clears them.

## Sentinel Convention (now uniform across all optional update fields)

| Field | Omit/null | Clear sentinel | Set |
|-------|-----------|---------------|-----|
| `tags` | preserve existing | `[]` | non-empty list |
| `confidence` | preserve existing | _(not clearable)_ | float |
| `source_url` | preserve existing | `""` | string |
| `derived_from_ids` | preserve existing | `[]` | non-empty list |
| `expires_at` | preserve existing | `""` | ISO string |

Fixes agent-lore/lithos#37

### DIFF
--- a/src/lithos/knowledge.py
+++ b/src/lithos/knowledge.py
@@ -844,13 +844,22 @@ class KnowledgeManager:
         agent: str,
         content: str | None = None,
         title: str | None = None,
-        tags: list[str] | None = None,
-        confidence: float | None = None,
+        tags: list[str] | _UnsetType = _UNSET,
+        confidence: float | _UnsetType = _UNSET,
         source_url: str | None | _UnsetType = _UNSET,
         derived_from_ids: list[str] | None | _UnsetType = _UNSET,
         expires_at: datetime | None | _UnsetType = _UNSET,
     ) -> WriteResult:
         """Update an existing document.
+
+        tags semantics:
+        - _UNSET (default): preserve existing tags
+        - []: clear all tags
+        - non-empty list: replace tags
+
+        confidence semantics:
+        - _UNSET (default): preserve existing confidence
+        - float: set new value
 
         source_url semantics:
         - _UNSET (default): preserve existing source_url, no map change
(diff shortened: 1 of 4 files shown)


hanumanclaw:

hanumanclaw (commented):

**⚠️ Merge order dependency:** This PR **must be merged before or immediately after #59**. PR #59 passes `_UNSET` to `knowledge.update()` for `tags` and `confidence` — those parameters only accept `_UnsetType` after this PR's signature changes land. Merging #59 onto a main that does not have #58 would cause a `TypeError` at runtime (data corruption risk via silent field erasure in the old signature).

- Formalises the `_UNSET` sentinel pattern for all optional update fields (`tags`, `confidence`, `source_url`, `derived_from_ids`, `expires_at`)
- Callers that omit a field now pass `_UNSET` (preserve existing value) rather than `None` (clear)
- Prevents accidental null-overwrites of existing metadata during partial updates

**No issues found.** The sentinel type is properly narrowed via `isinstance` checks before use. All call sites updated consistently.

davesnowdon (approved):

Issue #38, opened 2026-03-11
[MEDIUM] Slug map has silent last-writer-wins semantics


davesnowdon:
## Summary
Slug collisions are silently overwritten, potentially causing wiki-links to resolve to wrong documents.

## Audit Reference
- **Audit ID:** 2.4
- **Severity:** 🟡 Medium
- **Category:** Architecture

## Affected Files
- `src/lithos/knowledge.py`

## Problem
The slug-to-ID mapping uses simple assignment:
```python
_slug_to_id[slug] = doc_id
```

When two documents generate the same slug (e.g., "My Document" and "my-document"), the second silently overwrites the first. Wiki-links using `[[my-document]]` will resolve to whichever document was indexed last.

This can cause:
- Incorrect link resolution
- Data loss (links to original document become unreachable)
- Confusing behavior for agents relying on wiki-links

## Recommendation
1. Check for collisions in `create()` and reject with clear error
2. During `_scan_existing()`, log WARNING when collision detected
3. Consider appending suffix for disambiguation (e.g., `my-document-2`)
4. Add `lithos validate` check for slug collisions

## Additional Context
This is similar to URL slug collision handling in CMS systems. Most choose either reject or auto-suffix approach.

PR #59, opened 2026-03-11
fix: detect slug collisions, warn on scan, raise on create/update

hanumanclaw:
## Summary

Eliminates the silent last-writer-wins behaviour in the slug map. Collisions during startup scan now produce a warning; collisions during `create`/`update` raise an explicit `SlugCollisionError`.

## Problem

When two notes resolved to the same slug, the later-processed document silently overwrote the earlier one in `_slug_to_id`. This caused stale lookups and data loss without any indication.

## Changes

- `src/lithos/errors.py`: Added `SlugCollisionError(ValueError, LithosError)` with `slug` and `existing_id` attributes
- `src/lithos/knowledge.py`:
  - `_scan_existing()`: Before writing to `_slug_to_id`, check for a different existing owner. If found, log a `WARNING` and skip (first-seen-wins, deterministic since candidates are sorted). Does not raise — existing data may have collisions that need manual resolution.
  - `create()`: Before writing the slug, raise `SlugCollisionError` if the slug is already owned by a different document
  - `update()` (title-change path): Same guard before updating `_slug_to_id`
- `src/lithos/server.py`: Catch `SlugCollisionError` in `lithos_write` and return `{"status": "error", "code": "slug_collision", "message": ...}`
- Tests: Added tests for warning on scan collision, `SlugCollisionError` on create/update, and the server error envelope

Fixes agent-lore/lithos#38

### DIFF
--- a/src/lithos/knowledge.py
+++ b/src/lithos/knowledge.py
@@ -572,7 +572,7 @@ class KnowledgeManager:
                             )
                         else:
                             self._slug_to_id[slug] = doc_id
-                        self._id_to_title[doc_id] = title
+                            self._id_to_title[doc_id] = title
 
                     # Populate metadata cache for filtering
                     raw_updated = post.metadata.get("updated_at")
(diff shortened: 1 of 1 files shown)


hanumanclaw:

PR #60, opened 2026-03-11
feat: structured error types — SearchBackendError replaces silent return []

hanumanclaw:
## Summary

Addresses the core feedback on this branch: `full_text_search` and `semantic_search` were silently swallowing exceptions and returning `[]`, making it impossible for callers to distinguish *backend failure* from *no matching documents*.

## Design decision

A new `lithos/errors.py` module introduces a proper error hierarchy:

```python
class LithosError(Exception):
    """Base error for all Lithos operations."""

class SearchBackendError(LithosError):
    """One or more search backends failed."""
    def __init__(self, message: str, backend_errors: dict[str, Exception]): ...

class IndexingError(LithosError):
    """All backends failed during a write operation."""
    def __init__(self, message: str, backend_errors: dict[str, Exception]): ...
```

The `backend_errors` dict names the failing backend(s) (`"tantivy"`, `"chroma"`) and carries the original exception for full context.

## Behaviour changes

| Method | Before | After |
|---|---|---|
| `full_text_search` | catches exc, logs warning, returns `[]` | raises `SearchBackendError` — caller decides how to degrade |
| `semantic_search` | catches exc, logs warning, returns `[]` | raises `SearchBackendError` — caller decides how to degrade |
| `index_document` | swallows both backend failures silently | logs warning on partial failure; raises `IndexingError` if *both* fail |
| `remove_document` | swallows both backend failures silently | logs warning on partial failure; raises `IndexingError` if *both* fail |

## Tests

- Removed old tests that asserted `returns []` on error
- Added `pytest.raises(SearchBackendError)` tests for both search methods
- Added `pytest.raises(IndexingError)` tests for total failure in `index_document` and `remove_document`
- Added a test that `SearchBackendError` is a subclass of `LithosError` (broad-catch contract)
- Preserved partial-failure (one backend works) tests

### DIFF
--- a/src/lithos/errors.py
+++ b/src/lithos/errors.py
@@ -30,9 +30,7 @@ class SearchBackendError(LithosError):
         self.backend_errors: dict[str, Exception] = backend_errors
 
     def __str__(self) -> str:
-        detail = ", ".join(
-            f"{backend}: {exc}" for backend, exc in self.backend_errors.items()
-        )
+        detail = ", ".join(f"{backend}: {exc}" for backend, exc in self.backend_errors.items())
         return f"{super().__str__()} [{detail}]"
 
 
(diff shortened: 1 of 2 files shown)


hanumanclaw:
Closing as superseded. The upstream  has already incorporated all changes from this PR —  (with /), the error-raising logic in , and the  test suite are all present in current main. Nothing to add here.

Issue #40, opened 2026-03-11
[MEDIUM] Embedding model loaded synchronously at startup


davesnowdon:
## Summary
Sentence-transformers model loads synchronously, blocking the event loop for several seconds.

## Audit Reference
- **Audit ID:** 2.6
- **Severity:** 🟡 Medium
- **Category:** Architecture

## Affected Files
- `src/lithos/search.py`

## Problem
The `sentence-transformers` model loads during `SearchEngine.__init__()`. On first load (before caching), this:
1. Downloads model weights (if not cached)
2. Loads model into memory
3. Initializes CUDA/CPU backends

This blocks the async event loop for 2-5+ seconds, causing:
- Slow server startup
- Timeout issues for clients waiting for first response
- Poor UX when running `lithos serve`

## Recommendation
1. Lazy-initialize model on first semantic search request
2. Alternatively, load via `ThreadPoolExecutor` with awaited future
3. Add startup log message indicating model loading progress
4. Consider `LITHOS_PRELOAD_EMBEDDINGS=true` for deployments wanting eager load

## Additional Context
This is a common pattern with ML models. FastAPI/Starlette apps typically use lifespan events with thread pool for heavy initialization.

PR #63, opened 2026-03-11
fix: load embedding model asynchronously to unblock startup

hanumanclaw:
## Summary

The SentenceTransformer embedding model was loaded synchronously on the event loop thread when the first semantic search (or index rebuild) accessed `ChromaIndex.model`, blocking server startup.

## Changes

- `src/lithos/search.py` (`ChromaIndex`): Added `_model_lock` (lazily-initialised `asyncio.Lock`), `async ensure_model_loaded()` that loads the model via `asyncio.to_thread` under the lock (idempotent — only loads once), and `async embed_async()` for async embedding. Sync `model` property retained for backward compatibility.
- `src/lithos/search.py` (`SearchEngine`): Added `async ensure_embeddings_loaded()` delegating to `chroma.ensure_model_loaded()`.
- `src/lithos/server.py` (`initialize()`): Fires `asyncio.create_task(self.search.ensure_embeddings_loaded())` so the model pre-warms in the background without blocking the startup coroutine.
- `tests/`: Added `test_ensure_model_loaded_is_nonblocking` verifying concurrent calls load the model only once.

## Test Results

All 5 checks pass:
- `pytest -m "not integration"`: 366 passed
- `pytest -m integration`: 209 passed
- `ruff check`: clean
- `ruff format --check`: clean
- `pyright`: 0 errors

Fixes agent-lore/lithos#40

### DIFF
--- a/src/lithos/search.py
+++ b/src/lithos/search.py
@@ -10,7 +10,7 @@ import time
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import chromadb
 import tantivy
@@ -22,7 +22,9 @@ from lithos.knowledge import KnowledgeDocument
 from lithos.telemetry import lithos_metrics, traced
 
 if TYPE_CHECKING:
+    import numpy as np
     from chromadb.api import ClientAPI
+    from numpy.typing import NDArray
 
 logger = logging.getLogger(__name__)
 
@@ -504,7 +506,7 @@ class ChromaIndex:
             if self._model is None:
                 self._model = await asyncio.to_thread(SentenceTransformer, self.model_name)
 
-    async def embed_async(self, texts: list[str]) -> Any:
+    async def embed_async(self, texts: list[str]) -> NDArray[np.float32]:
         """Generate embeddings asynchronously without blocking the event loop.
 
         NOTE: This method is currently unused (dead code).  It is kept here as
--- a/src/lithos/server.py
+++ b/src/lithos/server.py
@@ -225,10 +225,19 @@ class LithosServer:
                 await self._rebuild_indices()
 
         # Pre-warm the embedding model in the background so the first real
-        # request does not block the event loop.
-        task = asyncio.create_task(self.search.ensure_embeddings_loaded())
-        self._background_tasks.add(task)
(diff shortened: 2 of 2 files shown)


hanumanclaw (commented):
## Argus Review — APPROVED ✅

**CI (all 5 verified locally)**
| Check | Result |
|---|---|
| pytest (not integration) | 366 passed ✅ |
| pytest (integration) | 209 passed ✅ |
| ruff check | clean ✅ |
| ruff format --check | clean ✅ |
| pyright | 0 errors ✅ |

---

### Is the async lock truly idempotent? Any race condition between concurrent first requests?

**No race condition. ✅**

The lock is created lazily:
```python
if self._model_lock is None:
    self._model_lock = asyncio.Lock()
async with self._model_lock:
    if self._model is None:
        self._model = await asyncio.to_thread(SentenceTransformer, self.model_name)
```

**Why this is safe:** asyncio is cooperative/single-threaded. Between the `if _model_lock is None:` check and the assignment, there is no `await`, so no other coroutine can interleave. Only one coroutine can create the lock. Once the lock exists, contending coroutines queue on `async with self._model_lock:` and the inner `if self._model is None:` double-check ensures the model is only loaded once. Classic and correct double-checked locking for asyncio. ✅

The test correctly validates this with `asyncio.gather()` under a 100ms simulated-slow constructor confirming `load_count == 1`.

---

### Observation — Background task exceptions are silently swallowed (non-blocking)

```python
task = asyncio.create_task(self.search.ensure_embeddings_loaded())
self._background_tasks.add(task)
task.add_done_callback(self._background_tasks.discard)
```

If `ensure_embeddings_loaded()` raises (e.g. model download fails, disk full, network error), the exception is discarded — Python will emit an `asyncio: Task exception was never retrieved` warning to stderr but nothing more. The server will continue running, and the next real request will fall through to the sync `model` property which attempts a blocking load on the event loop thread.

Suggest adding a second done callback that logs the exception:
```python
def _log_bg_task_exception(task: asyncio.Task) -> None:
    if not task.cancelled() and (exc := task.exception()):
        logger.error("Background task %s failed: %s", task.get_name(), exc)

task.add_done_callback(_log_bg_task_exception)
```

Non-blocking for this PR but worth a follow-up.

---

Overall: the async pattern is correct, idempotent, and race-free. The background task lifecycle management (`_background_tasks` set with discard callback) is solid. Correctly addresses #40.

hanumanclaw:

Issue #46, opened 2026-03-11
[MEDIUM] Wiki-link rename staleness is not auto-repaired


davesnowdon:
## Summary
Renamed documents leave broken wiki-links that accumulate over time.

## Audit Reference
- **Audit ID:** 4.2
- **Severity:** 🟡 Medium
- **Category:** Product/Usability

## Affected Files
- `src/lithos/knowledge.py`
- `src/lithos/graph.py`

## Problem
When a document is renamed:
1. Its slug changes
2. Other documents referencing it via `[[old-slug]]` now have broken links
3. `lithos validate --fix` doesn't rewrite the links
4. Links accumulate as dead over time

This degrades knowledge graph quality and agent experience.

## Recommendation
1. Auto-update wiki-links on rename via file watcher:
   - Detect slug change
   - Find all documents containing `[[old-slug]]`
   - Rewrite to `[[new-slug]]`
2. Gate behind `auto_repair_wiki_links` config flag (default: true)
3. Log repairs for auditability
4. Handle edge cases (circular renames, conflicting slugs)

## Additional Context
This is similar to how IDEs handle symbol renames. The file watcher already detects changes, just needs the repair logic.

PR #64, opened 2026-03-11
fix: detect and report stale wiki-links in reconcile graph scope

hanumanclaw:
## Summary

When a note is renamed, wiki-links in other notes that referenced it by the old title/slug went stale silently. The reconcile graph scope now detects these and includes them in the actions list as `stale_link` repair actions for operator review.

Note content is **never rewritten** — stale links are reported only.

## Changes

- `src/lithos/reconcile.py` (`_reconcile_graph`): After confirming the graph cache is consistent (no rebuild needed), calls `graph.get_broken_links()` and appends a `stale_link` action per unresolved link with `source_id`, `source_title`, `link_target`, and `reason: "target_slug_not_found"`. Both dry-run and real run report them without touching note files.
- `tests/test_reconcile.py`: Fixed `test_graph_real_run_idempotent` (second run now correctly produces `noop` by ensuring all links resolve). Added `TestReconcileStaleWikiLinks` with three tests: stale link reported in dry-run, stale link reported in real run with no content mutation, and no stale links when all links resolve.

## Test Results

All 5 checks pass:
- `pytest -m "not integration"`: 368 passed
- `pytest -m integration`: 209 passed
- `ruff check`: clean
- `ruff format --check`: clean
- `pyright`: 0 errors

Fixes agent-lore/lithos#46

### DIFF
--- a/src/lithos/reconcile.py
+++ b/src/lithos/reconcile.py
@@ -212,7 +212,14 @@ async def _reconcile_indices(config: LithosConfig, dry_run: bool) -> dict[str, A
 
 
 async def _reconcile_graph(config: LithosConfig, dry_run: bool) -> dict[str, Any]:
-    """Reconcile the wiki-link graph cache against the markdown corpus."""
+    """Reconcile the wiki-link graph cache against the markdown corpus.
+
+    When the cache is already consistent (no node/edge drift), a second pass
+    scans for stale wiki-links (targets that don't resolve to any document)
+    and reports them as ``stale_link`` actions.  Stale-link detection is
+    skipped when the cache itself needs a rebuild; the caller must reconcile
+    again after repair to surface stale links.
+    """
     tracer = get_tracer()
     actions: list[dict[str, Any]] = []
     failures: list[dict[str, Any]] = []


hanumanclaw (commented):
## Argus Review — APPROVED ✅

**CI (all 5 verified locally)**
| Check | Result |
|---|---|
| pytest (not integration) | 368 passed ✅ |
| pytest (integration) | 209 passed ✅ |
| ruff check | clean ✅ |
| ruff format --check | clean ✅ |
| pyright | 0 errors ✅ |

---

### Does reconcile NEVER rewrite note content?

**Confirmed: zero content writes. ✅**

The stale-link detection path in `_reconcile_graph` only builds a `stale_actions` list and returns it inside the `_make_result()` dict. No file I/O, no knowledge manager calls, no graph mutations. Both `dry_run=True` and `dry_run=False` take identical code paths — the `dry_run` flag is only threaded through `_make_result()` for metadata.

`test_stale_link_reported_in_real_run` correctly asserts `"[[NonExistentNote]]"` remains in the document after a real run. ✅

---

### Implementation sanity-checks

**`get_broken_links()` shape:** Returns `list[tuple[str, str, str]]` — `(source_id, source_title, link_target)` — which is correctly unpacked in reconcile. ✅

**`__unresolved__` sentinel pattern:** Edges pointing to nodes with `__unresolved__` prefix are the graph's existing representation for unresolved wiki-links. `get_broken_links()` correctly skips iterating over unresolved nodes as sources (they shouldn't have outgoing edges anyway) and detects outgoing edges from real nodes that target unresolved ones. ✅

---

### Observation — Stale links only reported when cache is already consistent (design note, non-blocking)

Stale link detection sits inside the `if not actions:` branch — it only runs after confirming the graph cache has no rebuild actions. If a first run finds cache inconsistencies, stale links are invisible until the following reconcile. This is a deliberate design choice and acceptable, but it means a freshly-renamed note produces a rebuild action on run 1 and a stale_link report on run 2. Operators should be aware they may need two reconcile passes after a rename to see the full picture.

---

### Regression fix in `test_graph_real_run_idempotent`

The addition of `await _create_doc(knowledge_manager, "Gamma Doc", "Gamma content.")` is correct and necessary: without a resolvable `Gamma Doc`, the `[[gamma-doc]]` link in `Delta Doc` would produce a stale_link action on the second run, breaking the `status == "noop"` assertion. Good fix. ✅

---

Overall: clean, targeted fix. No content mutation possible. Correctly addresses #46.

hanumanclaw (commented):

- Adds stale wiki-link detection to the reconcile graph scope
- Reports stale links (pointing to non-existent documents) during graph reconciliation without blocking writes
- Detection is scoped correctly to the reconcile pass — no impact on hot-path read/write operations

**No issues found.**

Issue #47, opened 2026-03-11
[MEDIUM] lithos_cache_lookup confidence parameter conflates meanings


davesnowdon:
## Summary
The `min_confidence` parameter conflates document reliability with retrieval quality.

## Audit Reference
- **Audit ID:** 4.3
- **Severity:** 🟡 Medium
- **Category:** Product/Usability

## Affected Files
- `src/lithos/server.py` - `lithos_cache_lookup` tool

## Problem
The `min_confidence` parameter filters on stored document confidence (a domain-level reliability score set by the writer), but callers often expect it to be a retrieval quality threshold.

Two different concepts are being conflated:
1. **Document confidence**: "How reliable is this information?" (set at write time)
2. **Match score**: "How well does this match my query?" (computed at search time)

Currently, semantic similarity scores from ChromaDB are not surfaced to callers.

## Recommendation
1. Rename stored field from `confidence` to `reliability` for clarity
2. Surface semantic similarity as `match_score` in response
3. Add separate `min_match_score` parameter for retrieval filtering
4. Document the distinction clearly in tool descriptions

## Additional Context
This naming confusion can lead to incorrect filtering. An agent might set `min_confidence=0.8` expecting high-quality matches, but instead filter out reliable documents that happen to have lower confidence scores.

PR #65, opened 2026-03-11
fix: separate confidence threshold and ranking params in lithos_cache_lookup

hanumanclaw:
## Summary

The `lithos_cache_lookup` `confidence` parameter conflated two distinct concepts in a single field, making the API ambiguous.

## Changes

- Retains `min_confidence: float = 0.5` as the **threshold filter** — docs below this score are excluded entirely (backward compatible)
- Adds new `sort_by_confidence: bool = False` parameter — when `True`, candidates that pass all filters are sorted by their `metadata.confidence` descending so the highest-confidence doc is returned rather than the first semantic match
- Updated docstring with explicit semantics for both parameters
- Added tests covering the new `sort_by_confidence` behaviour

## Backward compatibility

Default `sort_by_confidence=False` preserves existing behaviour exactly.

Fixes agent-lore/lithos#47

### DIFF
--- a/src/lithos/server.py
+++ b/src/lithos/server.py
@@ -876,7 +876,9 @@ class LithosServer:
                     is returned as the hit. When False (default), the first passing
                     candidate (in semantic-search order) is returned, preserving
                     backward-compatible behaviour.
-                limit: Max candidate docs to evaluate (default: 3)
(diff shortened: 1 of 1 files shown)


hanumanclaw:

Issue #41, opened 2026-03-11
[LOW] Claim expiry check has TOCTOU gap


davesnowdon:
## Summary
Task claim expiry check and new claim insert are separate SQL statements with race window.

## Audit Reference
- **Audit ID:** 2.7
- **Severity:** 🟢 Low
- **Category:** Architecture

## Affected Files
- `src/lithos/coordination.py`

## Problem
The claim acquisition logic:
1. SELECT to check if claim exists and is expired
2. INSERT new claim if check passes

Between steps 1 and 2, another agent could acquire the claim. The current code handles this (returns False on conflict), but the UX is retry-based rather than atomic takeover.

In high-contention scenarios, this could cause:
- Unnecessary retry loops
- Slightly unfair claim distribution
- Wasted round-trips

## Recommendation
1. Use single `INSERT OR REPLACE` with expiry condition in WHERE
2. Alternatively, wrap in `BEGIN EXCLUSIVE` transaction
3. Consider `INSERT ... ON CONFLICT DO UPDATE` for atomic upsert

## Additional Context
This is low severity because the current behavior is correct (no data corruption), just suboptimal. The retry approach works fine for typical Lithos workloads.

PR #67, opened 2026-03-11
fix: make claim expiry check and acquisition atomic, close TOCTOU gap

hanumanclaw:
## Summary

The `claim_task` method in `coordination.py` had a time-of-check/time-of-use (TOCTOU) race condition: it read the existing claim's expiry time and then acted on it in two separate SQL operations. Another agent could acquire the same claim between the SELECT and the subsequent UPDATE/INSERT.

## Changes

Replaced the SELECT + conditional UPDATE/INSERT pattern with a single atomic SQL statement:

```sql
INSERT INTO claims (task_id, agent, aspect, claimed_at, expires_at)
VALUES (?, ?, ?, ?, ?)
ON CONFLICT(task_id, aspect) DO UPDATE SET
    agent      = excluded.agent,
    claimed_at = excluded.claimed_at,
    expires_at = excluded.expires_at
WHERE claims.expires_at <= ? OR claims.agent = excluded.agent
```

The `WHERE` clause on the `DO UPDATE` fires only when:
- The existing claim is expired (`expires_at <= now`), **or**
- The same agent is renewing its own claim

When a different agent holds an active claim, the `WHERE` is false, the row is left unchanged, and `rowcount` returns 0 — all within a single database operation, eliminating the race window.

All existing behaviour is preserved (same agent renewal, different-agent blocking, expired claim takeover).

Fixes agent-lore/lithos#41

### DIFF
--- a/tests/test_coordination.py
+++ b/tests/test_coordination.py
@@ -487,6 +487,41 @@ class TestClaimManagement:
         upper_bound = datetime.now(timezone.utc) + timedelta(minutes=2)
         assert lower_bound < expires_at < upper_bound
 
+    @pytest.mark.asyncio
+    async def test_concurrent_claim_only_one_succeeds(
+        self, coordination_service: CoordinationService
+    ):
+        """When two coroutines race to claim the same task aspect, exactly one wins.
+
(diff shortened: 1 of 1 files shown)


hanumanclaw:

davesnowdon (approved):

PR #112, opened 2026-03-17
feat: hybrid lithos_search with RRF, remove lithos_semantic (fixes #76)

### LINKED ISSUE #76: [HIGH] Search/retrieval tool proliferation — consolidate before LCMA
## Problem

There are currently three retrieval tools (`lithos_search`, `lithos_semantic`, `lithos_cache_lookup`) with subtly different semantics. LCMA will add `lithos_retrieve`, bringing the total to four. An agent trying to find relevant knowledge has to choose between them, and the decision criteria are non-obvious.

| Tool | Backend | Returns content? | Freshness-aware? | Configurable ranking? |
|------|---------|-----------------|-------------------|----------------------|
| `lithos_search` | Tantivy | Snippets only | Yes (`is_stale`) | No |
| `lithos_semantic` | ChromaDB | Snippets only | Yes (`is_stale`) | No |
| `lithos_cache_lookup` | ChromaDB + URL lookup | Full content | Yes (core purpose) | Sort by confidence |
| `lithos_retrieve` (LCMA) | Multi-scout PTS | TBD | TBD | Temperature, terraces |

## Recommendation

- When `lithos_retrieve` ships, `lithos_search` and `lithos_semantic` should become internal scouts, not top-level MCP tools.
- For backward compat they can remain registered but descriptions should direct agents to prefer `lithos_retrieve`.
- Long-term: a single `lithos_search` with `mode` parameter (`fulltext | semantic | hybrid | auto`) is cleaner than two separate tools.
- `lithos_cache_lookup` is different enough to justify separate existence (it's a "should I research this?" workflow primitive).

## Pre-LCMA action

Merge `lithos_search` and `lithos_semantic` or add a `mode` param to avoid shipping 4 retrieval tools.

## Source

Tool surface analysis report — Finding #1

hanumanclaw:
## Summary

Implements [#76](https://github.com/agent-lore/lithos/issues/76) — removes `lithos_semantic` and upgrades `lithos_search` to hybrid-by-default with Reciprocal Rank Fusion.

### Breaking Change
- **`lithos_semantic` MCP tool removed.** Use `lithos_search` with `mode="semantic"` for pure semantic search, or the new default `mode="hybrid"` for best results.

### Added
- `lithos_search` gains a `mode` parameter: `hybrid` (default) | `fulltext` | `semantic`
- New `reciprocal_rank_fusion()` pure helper function in `search.py`
- New `SearchEngine.hybrid_search()` method — calls both Tantivy and ChromaDB, merges via RRF (k=60), deduplicates by doc ID, returns unified `SearchResult` list ranked by RRF score
- Telemetry/metrics updated to include `mode` label

### Mode behaviour
- `mode="fulltext"` → Tantivy only (existing behaviour, unchanged)  
- `mode="semantic"` → ChromaDB only (lithos_semantic's existing behaviour)  
- `mode="hybrid"` (default) → both backends, RRF merge, deduplicated by doc ID

### Tests
- All 46 tests pass
- New `TestHybridSearch` class covering: RRF pure function (k=60 math, common doc boost, empty lists), hybrid returns results, hybrid deduplicates by doc ID

Closes #76

### DIFF
--- a/src/lithos/server.py
+++ b/src/lithos/server.py
@@ -824,7 +824,7 @@ class LithosServer:
             author: str | None = None,
             path_prefix: str | None = None,
             threshold: float | None = None,
-        ) -> dict[str, list[dict[str, Any]]]:
+        ) -> dict[str, Any]:
             """Search across the knowledge base.
 
             Supports three search modes:


hanumanclaw (commented):
## Argus Review 🔍

**Status:** CHANGES REQUESTED _(posted as comment — cannot request changes on own PR)_

### ✅ Passed
- All 46 unit tests pass (`tests/test_search.py`)
- Scope matches issue #76 — `lithos_semantic` removed, `lithos_search` upgraded to hybrid-by-default with RRF
- No security concerns (no eval/exec/subprocess/hardcoded secrets)
- Test updates in `test_integration_conformance.py` are clean and consistent — all 6 `lithos_semantic` call sites correctly migrated to `lithos_search(mode="semantic")`
- `reciprocal_rank_fusion()` implementation is correct and well-tested
- Single-backend fault tolerance in `hybrid_search()` is a good resilience choice
- CHANGELOG entries are accurate and complete
- PR description matches what actually changed

---

### ❌ Issues (blocking)

**[Missing SHOWBOAT.md / DEMO.md]**
No executable proof-of-work document is present in this PR. Per review policy, a `SHOWBOAT.md` or `DEMO.md` with real `exec` output blocks is required before merge. This verifies the new `hybrid` mode actually runs end-to-end (both backends fire, RRF scores are reasonable, deduplication works) rather than just passing unit tests with mocks.

---

### ⚠️ Concerns (non-blocking)

**[src/lithos/server.py ~L892 / src/lithos/search.py L910] `author`/`path_prefix` silently ignored in `mode="semantic"`**

`ChromaIndex.search()` was correctly updated to accept `author` and `path_prefix` (post-query filtering). `SearchEngine.hybrid_search()` wires them through to Chroma. But `SearchEngine.semantic_search()` (line 910) was *not* updated — its signature has no `author`/`path_prefix` params, so the server's `mode="semantic"` path drops them silently:

```python
# server.py ~L892 — author and path_prefix are in scope but never forwarded
elif mode == "semantic":
    sem_results = self.search.semantic_search(
        query=query, limit=limit, threshold=threshold, tags=tags
        # author= ?   path_prefix= ?
    )
```

A caller doing `lithos_search(mode="semantic", author="alice")` will get back results for all authors with no warning. Two acceptable fixes:
1. **Propagate:** update `SearchEngine.semantic_search()` to accept and forward them (consistent with hybrid path — the capability already exists in `ChromaIndex.search()`)
2. **Reject:** raise a `ValueError` in server.py if `author`/`path_prefix` are provided with `mode="semantic"` (docstring currently says "fulltext/hybrid only" but doesn't enforce it)

Option 1 is cleaner.

**[tests/test_search.py] No direct filter tests for new `ChromaIndex` params**

`TestHybridSearch` tests the pipeline end-to-end but there are no unit tests specifically asserting that `author=` and `path_prefix=` filtering in `ChromaIndex.search()` actually excludes the right docs. Low-risk given the implementation is straightforward, but worth a dedicated test given this was the stated bug fix.

---

### 📋 Checklist
- [ ] Showboat doc present and verified ← **BLOCKING**
- [x] Tests pass (46/46)
- [x] Scope matches issue #76
- [x] No security concerns
- [x] PR description accurate

---
*— Argus 🔍 (automated review on behalf of hanumanclaw)*

hanumanclaw:
## Re-review: APPROVED ✅

**Argus (code-reviewer subagent) — second pass on feat/hybrid-search-rrf**

All three originally requested changes are correctly addressed. Here's a full assessment of the updated state.

---

### ✅ Previously requested changes — all resolved

**1. SHOWBOAT.md** — Present and well-executed. Shows real output for all three modes, the RRF reranking effect, and the author-filter fix with zero cross-contamination. The absolute path in the `uv run` header (`file:///Users/hanuman/...`) makes the demo non-portable, but for a proof-of-work document that's an acceptable trade-off.

**2. Silent filter bug fixed** — `SearchEngine.semantic_search()` now correctly passes `author` and `path_prefix` through to `ChromaIndex.search()`. Confirmed in `search.py`. `hybrid_search()` also passes both params to both backends, so the fix extends to hybrid mode too.

**3. Unit tests — `TestChromaIndexFilters` (6 tests)** — All six tests are well-structured and test the right things: author inclusion, author exclusion, path prefix inclusion, combined AND semantics, and the two engine-level wiring tests. Good use of `threshold=0.0` to remove score-cutoff noise from filter tests.

**`TestHybridSearch` (4 tests)** — RRF maths test is solid (k=60 formula check, common-doc boost, empty lists). Dedup and basic results tests cover the core contract.

---

### 💬 Minor comments (non-blocking)

**`test_hybrid_deduplicates_by_doc_id` — assertion is too weak:**
```python
assert ids.count(doc.id) <= 1  # passes even if doc is absent
```
This passes vacuously if the doc isn't returned at all. Should be:
```python
assert doc.id in ids
assert ids.count(doc.id) == 1
```

**`lithos_search` raises `ValueError` for unknown mode instead of returning an error envelope:**
```python
raise ValueError(f"Unknown search mode {mode!r}. ...")
```
Every other validation error in this server returns `{"status": "error", "code": ..., "message": ...}`. FastMCP will propagate a raw exception as an MCP-level error, which is technically fine, but it breaks the consistent error contract callers expect. Consider returning a structured envelope here too. The CHANGELOG explicitly calls this out as intentional, so leaving it to the author's discretion.

**Filter coverage gap in `TestHybridSearch`:** No test confirms that `hybrid_search(author=...)` correctly filters results. The code clearly wires it through, but a test would lock in that guarantee — especially since a silent-filter regression is exactly what this PR was fixing.

---

### Overall

Core implementation is correct. RRF math is right. The filter bug is genuinely fixed and properly tested at both the `ChromaIndex` and `SearchEngine` levels. Integration conformance tests updated cleanly. CHANGELOG is thorough and honest about the breaking changes.

The dedup assertion nit is the only thing I'd encourage fixing before the next PR touches these tests. Everything else can live in a follow-up.

**Clean pass. Ready to merge.** 🚀

*— Argus, code-reviewer subagent*

PR #126, opened 2026-03-25
feat: LCMA prerequisites — async search wrapping, merge_and_normalize, edge.upserted event

hanumanclaw:
## Summary

Four prerequisite items from `docs/plans/lcma-checklist.md`, required before MVP 1 implementation begins. All 603 existing tests pass.

---

## 1. Wrap sync search methods in `asyncio.to_thread()`

`server.py` had three hot-path call sites where sync `SearchManager` methods were called directly inside async tool handlers:

| Tool | Method | Location |
|---|---|---|
| `lithos_search` | `full_text_search()`, `semantic_search()`, `hybrid_search()` | mode dispatch block |
| `lithos_cache_lookup` | `semantic_search()` | fallback path |
| `lithos_list` | `full_text_search()` | `content_query` filter (had a TODO comment) |

All three methods are plain `def` on `SearchManager` (tantivy + chromadb blocking I/O). Wrapping them at the call site prevents the event loop from stalling when LCMA adds 7+ parallel scouts in MVP 1.

`SearchManager` itself is not modified.

---

## 2. `merge_and_normalize()` in `src/lithos/lcma/utils.py`

New package `src/lithos/lcma/` with:

- **`Candidate`** dataclass: `node_id`, `score`, `reasons`, `scouts`
- **`merge_and_normalize(candidates)`** implementing design doc §5.3.1:
  - Per-scout min-max normalisation to `[0, 1]`
  - Single-element / all-equal-score groups normalise to `1.0`
  - Merge by `node_id`: max score wins, reasons concat, scouts union
  - Input is **not mutated**

**16 tests** in `tests/test_lcma_utils.py` covering:
- Empty input
- Single scout (single candidate, multiple candidates, equal scores)
- Multiple scouts with different raw score ranges (BM25 vs cosine vs RRF)
- Duplicate `node_id` across scouts (max score, reason union, scout dedup)
- Input immutability

---

## 3. `EDGE_UPSERTED = "edge.upserted"` event type

Added to `src/lithos/events.py` alongside existing constants (`NOTE_CREATED`, `FINDING_POSTED`, etc.). Required by the enrich worker event subscription and the `lithos_edge_upsert` tool (both MVP 1+).

---

## 4. `ENRICH_SUBSCRIBER_QUEUE_SIZE = 10_000` constant

Added to `src/lithos/events.py` with explanatory comment. The default `EventBus` subscriber queue is `100` and drops silently under load (design doc §8.10). The enrich worker will pass this constant to `EventBus.subscribe()` when it is wired up in MVP 2 — no worker implementation in this PR.

---

## Test results

```
603 passed, 1 warning in 373.41s
```

The one warning is a pre-existing `datetime.utcnow()` deprecation in `test_knowledge.py` — not introduced by this PR.

### DIFF
--- a/src/lithos/lcma/utils.py
+++ b/src/lithos/lcma/utils.py
@@ -62,11 +62,12 @@ def merge_and_normalize(candidates: list[Candidate]) -> list[Candidate]:
     # Multi-scout Candidates are produced as *output* of this function, not
     # accepted as input (they would corrupt the per-scout normalisation step).
     for c in candidates:
-        assert len(c.scouts) <= 1, (
-            f"merge_and_normalize expects single-scout Candidates as input; "
-            f"got scouts={c.scouts!r} for node_id={c.node_id!r}. "
-            "Multi-scout Candidates are produced as output, not accepted as input."
-        )
+        if len(c.scouts) > 1:
+            raise ValueError(
+                f"merge_and_normalize expects single-scout Candidates as input; "
+                f"got scouts={c.scouts!r} for node_id={c.node_id!r}. "
+                "Multi-scout Candidates are produced as output, not accepted as input."
+            )
 
     # --- Step 1: group by scout name ---
     by_scout: dict[str, list[Candidate]] = {}


src/lithos/lcma/utils.py
```diff
@@ -0,0 +1,118 @@
...
+        The caller is responsible for any final ranking/sorting.
+    """
+    if not candidates:
+        return []
+
+    # Guard: each input Candidate must come from exactly one scout.
+    # Multi-scout Candidates are produced as *output* of this function, not
+    # accepted as input (they would corrupt the per-scout normalisation step).
+    for c in candidates:
+        assert len(c.scouts) <= 1, (
```
davesnowdon: I don't think we should be using assert in non-test code

PR #170, opened 2026-04-14
LCMA MVP2

davesnowdon:
Closes #133


src/lithos/lcma/enrich.py
```diff
@@ -0,0 +1,705 @@
...
+# --- Entity extraction patterns ---
+# [[wiki-link]] or [[wiki-link|display]]
+_WIKI_LINK_RE = re.compile(r"\[\[([^\]\[|]+?)(?:\|[^\]]+)?\]\]")
+# `backtick-enclosed terms` (single backtick, not code fences)
+_BACKTICK_RE = re.compile(r"(?<!`)``?([^`\n]+?)``?(?!`)")
+# Capitalized multi-word phrases (2+ words starting with uppercase)
+_CAP_PHRASE_RE = re.compile(r"\b([A-Z][a-z]+(?:\s+[A-Z][a-z]+)+)\b")
+# Single capitalized words: proper nouns and PascalCase identifiers (3+ chars)
+_PROPER_NOUN_RE = re.compile(r"(?<!\w)([A-Z][a-zA-Z]{2,})(?!\w)")
+# Common English words to exclude from proper noun extraction
```
davesnowdon, line 47: This is a bit nasty. This should be separated out. There must be libraries that can do this for us

src/lithos/lcma/enrich.py
```diff
@@ -0,0 +1,705 @@
...
+    FINDING_POSTED,
+    EDGE_UPSERTED,
+]
+
+
+_CODE_FENCE_RE = re.compile(r"```[\s\S]*?```")
+
+
+def _extract_entities_from_text(text: str) -> list[str]:
+    """Extract entity names from text using rule-based heuristics.
```
davesnowdon, line 209: We should look a using a library to do named entity recognition properly

src/lithos/lcma/enrich.py
```diff
@@ -0,0 +1,705 @@
...
+                await asyncio.sleep(interval)
+                try:
+                    await self.drain()
+                except Exception:
+                    logger.exception("EnrichWorker: drain cycle failed")
+        except asyncio.CancelledError:
+            return
+
+    async def drain(self) -> None:
+        """Process pending nodes and tasks from enrich_queue."""
```
davesnowdon, line 426: There are many places like this where it would be good to have some logging

src/lithos/config.py
```diff
@@ -15,16 +15,26 @@
...
+    "vector": 0.25,
+    "lexical": 0.18,
+    "exact_alias": 0.10,
+    "tags_recency": 0.07,
+    "freshness": 0.04,
+    "provenance": 0.04,
+    "task_context": 0.04,
+    "graph": 0.13,
+    "coactivation": 0.10,
+    "source_url": 0.05,
```
davesnowdon, line 27: how were these generated?

src/lithos/config.py
```diff
@@ -15,16 +15,26 @@
...
 }
 
-_DEFAULT_NOTE_TYPE_PRIORS: dict[str, float] = {nt: 0.5 for nt in sorted(_LCMA_NOTE_TYPES)}
+_DEFAULT_NOTE_TYPE_PRIORS: dict[str, float] = {
+    "agent_finding": 0.6,
+    "concept": 0.45,
+    "hypothesis": 0.5,
+    "observation": 0.5,
+    "summary": 0.55,
+    "task_record": 0.35,
```
davesnowdon, line 36: how were these generated?

PR #246, opened 2026-05-08
refactor(intake): CorpusIntake module — route lithos_delete through it (PR 1 of 2)

davesnowdon:
## Summary

Introduces a `CorpusIntake` module that owns the five-step Corpus-mutation pipeline previously inlined inside MCP handlers. Routes `lithos_delete` through it; `lithos_write` follows in PR 2.

The pipeline (ensure-agent → call `KnowledgeManager` → sync Search engine → sync link graph → emit `NOTE_*`) lived as repeated inline blocks with unwritten ordering contracts (event-after-indices, no lock outside `KnowledgeManager._write_lock`, search-failure leaves doc on disk and skips the event). It is now a named seam.

This complements:
- **ADR-0001** (Reconcile lives on `KnowledgeManager`): intake is the agent-driven counterpart to corpus-driven Reconcile.
- **ADR-0002** (Search engine hides Tantivy/Chroma): intake calls only the public `SearchEngine` surface.

See [docs/adr/0003-corpus-intake.md](docs/adr/0003-corpus-intake.md) for the design decision and rejected alternatives, and [CONTEXT.md](CONTEXT.md) for the new glossary entry.

## What's in this PR

- `src/lithos/intake.py` — `CorpusIntake` class with `delete()` plus `DeleteRequest` / `DeleteOutcome` frozen dataclasses
- `src/lithos/server.py` — constructs intake in `initialize()` after `SearchEngine.create()`; `lithos_delete` collapses to a 7-line wrapper
- `CONTEXT.md` — adds **Corpus intake** term and the agent-driven-vs-corpus-driven relationship
- `docs/adr/0003-corpus-intake.md` — accepted ADR
- `tests/test_intake.py` — 6 unit tests + 2 integration tests pinning ordering and failure-mode invariants

## Load-bearing invariants preserved

1. **Atomic version check** — intake never reads-then-writes; `KnowledgeManager._write_lock` is the only serialisation point.
2. **Event-after-indices** — `await search.remove` → `graph.remove_document` (debounces) → `await emit`.
3. **No lock at intake** — confirmed by `test_delete_search_failure_propagates_with_no_event`.
4. **Emit failure doesn't undo writes** — confirmed by `test_delete_event_emit_failure_does_not_undo_corpus_delete`.
5. **Search/graph failures propagate, no event** — Drift is Reconcile's job per ADR-0001; same observable behaviour as today.
6. **`ensure_agent_known` runs before any mutation, even on no-op deletes** — confirmed by `test_delete_ensures_agent_before_corpus_mutation` and `test_delete_returns_not_found_for_unknown_id`.

## Out of scope

- `lithos_write` routing (deferred to PR 2 — separate issue with full implementation brief).
- File-watcher path at `server.py:_on_file_event` runs a similar but distinct pipeline (event-first because cache eviction happens during `KnowledgeManager.delete`). Not folded into intake; ADR-0003 explains why.
- Task-lifecycle / projection edge / read-only tools — different shapes, different homes.

## Test plan

- [x] `make lint` — clean
- [x] `make typecheck` — 0 errors, 0 warnings
- [x] `make test` — 1080 unit tests passed
- [x] `tests/test_intake.py` — 8/8 passed
- [x] `tests/test_server.py::TestErrorEnvelopes` (delete subset) — 8/8 passed
- [x] `tests/test_server.py::TestDeleteAgentRequired` — 2/2 passed
- [x] `tests/test_event_emission.py` — 14/14 passed
- [ ] CI green on this branch

> Note: full `make test-integration` on the dev host hits `OSError: [Errno 24] inotify instance limit reached` (kernel cap is 128 on this host); this is environmental and reproduces on `main` for the same subsets. CI should not hit this limit.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -28,11 +28,16 @@ _Avoid_: diff, repair list.
 The module owning every full-text and semantic search concern. Holds the Tantivy and Chroma indexes internally; their existence is not part of its interface. Health is one signal — agents either get a healthy engine or an unhealthy one with a reason; subsystem-level diagnostics are operator concerns.
 _Avoid_: search service, indexer.
 
+**Corpus intake**:
+The controlled entry point for **Corpus** mutations from agent tools. Ensures agent registration, runs the mutation through the corpus manager (which provides atomicity, including `expected_version` checks), synchronises derived views (Search engine, then link graph), and emits the matching `NOTE_*` event. Distinct from **Reconcile**: intake is agent-driven and updates views as a write happens; Reconcile is corpus-driven and brings views back into agreement after **Drift**.
+_Avoid_: ingestion, writer, pipeline, mutator.
+
 ## Relationships
 
 - The **Corpus** is the source of truth. The **Search engine**, the link graph, and the provenance projection are derived views.
 - A view is in **Drift** when its contents disagree with the **Corpus**.
 - A **Reconcile** consumes the **Corpus**, produces a **Reconcile plan** describing the **Drift**, then applies it to bring the view into agreement.
+- A **Corpus intake** is the inverse direction: an agent-driven mutation flows from the tool surface, through the corpus, out to every derived view, and onto the event bus.
 - The **Search engine** consumes **Indexable documents**, never `KnowledgeDocument` directly.
 
 ## Example dialogue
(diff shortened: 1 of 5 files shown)


davesnowdon:
PR 2 (route lithos_write through intake) tracked at #247 — fully specified, ready-for-agent.

Issue #262, opened 2026-05-09
ADR-0005F: Remove direct lcma imports from server.py + add import-graph guard test


davesnowdon:
## What to build

Final cleanup slice for ADR-0005. Remove all direct `from lithos.lcma.*` imports from `server.py` (they should now all flow through `self.memory.<method>` or `self.projection.<method>` after #257, #258, #259, #260 land). Add an import-graph guard test (or a ruff custom rule) that fails if any module outside `lithos.provenance` and `lithos.cognitive_memory` imports from `lithos.lcma.*`.

This is the slice that locks the seam. Without it, future PRs may quietly re-leak by importing `lithos.lcma.stats` somewhere new and the boundary erodes.

References: `docs/adr/0005-cognitive-memory-module.md` (`### Guardrail`).

## Acceptance criteria

- [ ] `grep -r 'from lithos.lcma' src/lithos/server.py` returns no matches (after the per-method migration slices).
- [ ] A guard test (in `tests/test_module_boundaries.py` or similar) fails if any file under `src/lithos/` other than `provenance.py` and `cognitive_memory.py` imports from `lithos.lcma.*`. Equivalent ruff custom rule is acceptable.
- [ ] The guard test is run as part of `make check`.
- [ ] CLI (`cli.py`) and any other top-level callers also stop importing `lithos.lcma.*` directly.
- [ ] `make check` and `make test-integration` are green.

## Blocked by

- #257 (retrieve migration)
- #258 (node_stats + edges migration)
- #259 (reinforcement migration)
- #260 (cache_lookup migration)

---

**Type**: AFK

davesnowdon:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Remove the server's direct LCMA imports and add an automated guard that prevents re-leakage.

**Current behavior:**
After #257-#260 land, all LCMA-internal imports in the server should already be gone — but nothing prevents a future change from re-introducing them. The architectural seam has no enforcement.

**Desired behavior:**
The server module imports nothing from the LCMA implementation directory — only from the **Provenance projection** and **Cognitive memory** facades. An automated guard (a unit test, a ruff custom rule, or an equivalent check) fails the build if any module outside those two facades imports from the LCMA implementation directory. The guard runs as part of `make check`.

**Key interfaces:**
- A guard test (or rule) that scans the source tree and fails if any file other than the two facade modules imports the LCMA implementation directory.
[...]

Issue #260, opened 2026-05-09
ADR-0005E: Migrate cache_lookup (and conflict_resolve if present) to CognitiveMemory


davesnowdon:
## What to build

Move the `lithos_cache_lookup` MCP tool implementation into `CognitiveMemory` as `CognitiveMemory.cache_lookup(...)`. If `lithos_conflict_resolve` exists today (or is being added in MVP2 per the LCMA roadmap), do the same: `CognitiveMemory.conflict_resolve(...)`.

Tool handlers in `server.py` collapse to one-line wrappers.

References: `docs/adr/0005-cognitive-memory-module.md` (`### Public surface`).

## Acceptance criteria

- [ ] `CognitiveMemory.cache_lookup(query) -> CachedResult` exists; `lithos_cache_lookup` handler is a one-line wrapper.
- [ ] If `lithos_conflict_resolve` exists today: `CognitiveMemory.conflict_resolve(...)` exists and the handler is a one-line wrapper. Otherwise: defer this part of the slice and note it for the MVP2 work.
- [ ] No direct LCMA internal imports for these methods in `server.py`.
- [ ] Tests cover the migrated methods through the Module's interface.
- [ ] `make check` and `make test-integration` are green.

## Blocked by

- #255 (CognitiveMemory scaffold).

---

**Type**: AFK

davesnowdon:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Move `lithos_cache_lookup` (and `lithos_conflict_resolve` if it exists today) onto **Cognitive memory**.

**Current behavior:**
`lithos_cache_lookup` is implemented inline in the server module or via a free function in the LCMA implementation. `lithos_conflict_resolve` may not exist yet (it's MVP2 work); if it does, same shape.

**Desired behavior:**
**Cognitive memory** exposes `cache_lookup` as a method. If `conflict_resolve` exists today, it does too. MCP handlers become one-line wrappers. If `conflict_resolve` does not exist, defer that part of the slice and note it on whichever MVP2 issue introduces the tool.

**Key interfaces:**
- `CognitiveMemory.cache_lookup(query) -> CachedResult` — mirrors today's `lithos_cache_lookup` tool signature.
- `CognitiveMemory.conflict_resolve(...)` — only if the tool exists today.
- MCP handlers for these tools import nothing from LCMA internals.

**Acceptance criteria:**
[...]

Issue #257, opened 2026-05-09
ADR-0005B: Migrate retrieve to CognitiveMemory + add error taxonomy


davesnowdon:
## What to build

Move the `lithos_retrieve` orchestration into `CognitiveMemory`. The `run_retrieve` function from `lcma/retrieve.py` becomes the implementation of `CognitiveMemory.retrieve(...)`; the MCP tool handler in `server.py` collapses to a one-line wrapper that calls `self.memory.retrieve(...)` and shapes the envelope.

Alongside, add the cognitive memory error taxonomy to `errors.py`:

- `CognitiveMemoryError(LithosError)` — base
- `ScoutFailure` — raised when a scout's backend fails (currently swallowed with a log warning in `retrieve.py:388–407`)
- Any timeout/retry types the retrieve path needs

`CognitiveMemory.retrieve` documents which exceptions propagate vs which are caught-and-logged. One contract, one place.

References: `docs/adr/0005-cognitive-memory-module.md` (`### Errors` section).

## Acceptance criteria

- [ ] `CognitiveMemory.retrieve(query, top_k, scopes, ...) -> RetrieveResult` exists and wraps `run_retrieve`.
- [ ] The `lithos_retrieve` MCP tool handler is a one-line wrapper around `self.memory.retrieve(...)` plus envelope shaping.
- [ ] `errors.py` defines `CognitiveMemoryError` and `ScoutFailure`; scout failures that today are swallowed with a log warning are routed through `ScoutFailure` (caught at the appropriate boundary, contract documented).
- [ ] Tests in `tests/test_cognitive_memory.py` cover retrieve through the seam, including scout failure handling.
- [ ] No direct `from lithos.lcma.retrieve` import in `server.py`.
- [ ] `make check` and `make test-integration` are green.

## Blocked by

- #255 (CognitiveMemory scaffold).

---

**Type**: AFK

davesnowdon:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Move the `lithos_retrieve` orchestration onto **Cognitive memory** and define a small error taxonomy for the retrieve path.

**Current behavior:**
`lithos_retrieve` runs through a free-function orchestrator that takes decomposed dependencies (multiple stores, the **Provenance projection**, the corpus manager). The MCP tool handler in the server constructs the call and shapes the envelope. Scout failures inside the orchestrator are caught and logged as warnings — the contract for "what can fail and how" is implicit.

[...]

Issue #251, opened 2026-05-09
ADR-0004A: ProvenanceProjection Module scaffold + scout-read migration


davesnowdon:
## What to build

Stand up `ProvenanceProjection` as a Module facade in `src/lithos/provenance.py` per ADR-0004 (https://github.com/agent-lore/lithos/blob/main/docs/adr/0004-provenance-projection-module.md). This slice delivers the scaffold and the public read surface only — no plan/apply yet, so it can land in parallel with the graph step (ADR-0001 step 2).

The Module wraps `EdgeStore` and exposes a public read API (`list_edges`, plus the smaller methods scouts use today). `EdgeStore` and `_project_provenance_to_edges` move to package-internal under `lcma/edges.py` (still in the file, but stop being imported from anywhere outside `provenance.py`). Scouts that today construct their own `EdgeStore` (notably `scout_provenance`, `scout_graph`, etc.) get migrated to read through `projection.list_edges(...)`.

Construction follows ADR-0002's eager-init pattern: `async def create(config) -> ProvenanceProjection`.

## Acceptance criteria

- [ ] `src/lithos/provenance.py` exists with `ProvenanceProjection`, `async create(config)`, public read methods, and `async close()`.
- [ ] All scouts read edges through `projection.list_edges(...)` (or equivalent public method); no `from lithos.lcma.edges import EdgeStore` outside `provenance.py`.
- [ ] `LithosServer.initialize()` constructs the projection via `await ProvenanceProjection.create(config)` and stores it as `self.projection`.
- [ ] Unit tests in `tests/test_provenance_projection.py` cover construction, lifecycle, and read-API behaviour.
- [ ] `make check` and `make test-integration` are green.

## Blocked by

None — can start immediately.

---

**Type**: AFK

davesnowdon:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Stand up the **Provenance projection** as its own Module facade; hide the edge store and projection function as package-internal; migrate scout reads through the new public read API.

**Current behavior:**
The edge store and the corpus-to-edges projection function live inside the **Cognitive memory** implementation directory and are imported directly by callers (the standalone reconcile module, scouts). There is no Module-level interface around them — every reader knows about the store class and constructs its own instance.

**Desired behavior:**
[...]

PR #270, opened 2026-05-12
feat(cognitive-memory): Module scaffold + lifecycle (#255)

### LINKED ISSUE #255: ADR-0005A: CognitiveMemory Module scaffold + lifecycle
## What to build

Stand up `CognitiveMemory` as a Module facade in `src/lithos/cognitive_memory.py` per ADR-0005. This slice delivers the scaffold and lifecycle only — no MCP tool methods migrated yet. `server.py` constructs the Module and starts/stops it but continues to call LCMA internals directly for retrieve, edges, stats, reinforcement.

The Module's constructor follows ADR-0002's eager-init pattern and takes `(config, knowledge, search, graph, projection, event_bus)`. `start()` opens the internal `StatsStore` and starts the `EnrichWorker`. `stop()` reverses lifecycle. `LithosServer.initialize()` constructs the Module after `ProvenanceProjection.create()` returns; `shutdown()` stops the Module first.

The point is to land the seam and lifecycle ordering before migrating any methods, so subsequent slices (#6, #7, #8, #9) can move methods one at a time without re-deciding constructor shape.

References: `docs/adr/0005-cognitive-memory-module.md`.

## Acceptance criteria

- [ ] `src/lithos/cognitive_memory.py` exists with `CognitiveMemory` and `async create(config, knowledge, search, graph, projection, event_bus)`.
- [ ] `async start()` opens `StatsStore` and starts the `EnrichWorker`. `async stop()` reverses cleanly without leaking threads.
- [ ] `LithosServer.initialize()` constructs the Module via `await CognitiveMemory.create(...)` and stores as `self.memory`. `LithosServer.shutdown()` calls `await self.memory.stop()` before closing other stores.
- [ ] Existing MCP tool handlers continue to work unchanged (still calling LCMA internals directly — migration in subsequent slices).
- [ ] Unit test: Module constructs, starts, stops without leaking threads.
- [ ] `make check` and `make test-integration` are green.

## Blocked by

- #251 (ADR-0004A: ProvenanceProjection scaffold) — the Module's constructor takes `projection: ProvenanceProjection`.

---

**Type**: AFK

davesnowdon:
## Summary

Stand up `CognitiveMemory` as a Module facade per [ADR-0005](docs/adr/0005-cognitive-memory-module.md). This is the **scaffold + lifecycle slice only** — no MCP tool handlers move. Method migration (`retrieve`, `edge_*`, `reinforce_*`, `node_stats`, `cache_lookup`) is in subsequent slices (#257-#260).

Closes #255.

## What changes

- **`src/lithos/cognitive_memory.py` (NEW)** — `CognitiveMemory` class with private `__init__`, async `create()`, `start()`, `stop()`. Owns `StatsStore` and `EnrichWorker` lifecycle.
- **`src/lithos/server.py`** — `LithosServer.__init__` drops the eager `StatsStore` construction; `initialize()` builds the Module via `await CognitiveMemory.create(...)` after `ProvenanceProjection.create(...)` returns; `self.stats_store` and `self._enrich_worker` are aliased from the Module so un-migrated handlers keep working unchanged (same pattern as `self.edge_store` aliasing `self.projection._edge_store` from #251). `stop_enrich_worker()` runs `await self.memory.stop()` before closing the projection.
- **`tests/test_cognitive_memory.py` (NEW)** — 9 unit tests covering construction (deps wired, StatsStore not yet opened), lifecycle (start opens store + starts worker, stop cleans up), idempotence (stop without start, double-stop, restart after stop), double-start error, and the LCMA-disabled path (start opens store but no worker).

## Design notes

ADR-0005 and the issue both list six dependencies for the Module. `EnrichWorker.__init__` requires `coordination: CoordinationService`, so the Module takes a **seventh** `coordination` argument. The alternative (server pre-builds `EnrichWorker` and passes it in) would conflict with the ADR's "Module owns the worker" stance.

## Verification

- [x] `make lint` — green
- [x] `make typecheck` — green (pyright standard mode, 0 errors)
- [x] `make test` — green (1139 unit tests pass)
- [x] `tests/test_cognitive_memory.py` covers Module lifecycle including no-thread-leak guarantee
- [x] Manual: `test_provenance_conformance.py` (21 tests) and `test_server.py` integration (115 tests) pass on this branch when run in isolation
- [x] Manual: combined `test_provenance_conformance.py + test_server.py` (136 tests) all pass on this branch

### Integration suite caveat

`make test-integration` shows pre-existing flakiness on `main` (verified by stashing this PR's changes locally: 93 errors before this branch, mostly cascading from `_NoOpSpan.set_*` TypeErrors in `test_smoke.py` / `test_telemetry.py`). With this branch the failure count shifts but the pattern is the same — setup-phase errors on the `server` fixture cascading from an unrelated earlier failure. Every integration test file passes individually on this branch.

CI will run the full suite — if anything in this branch genuinely breaks integration beyond the existing flake, the failure pattern will differ from main and we'll iterate.

## Acceptance criteria from #255

- [x] `src/lithos/cognitive_memory.py` exists with `CognitiveMemory` and `async create(config, knowledge, search, graph, projection, event_bus, coordination)` (extra `coordination` dep documented above).
- [x] `async start()` opens `StatsStore` and starts the `EnrichWorker`. `async stop()` reverses cleanly without leaking threads.
- [x] `LithosServer.initialize()` constructs the Module via `await CognitiveMemory.create(...)` and stores as `self.memory`. `LithosServer.shutdown()` calls `await self.memory.stop()` before closing other stores.
- [x] Existing MCP tool handlers continue to work unchanged (still calling LCMA internals directly via the server-level aliases).
- [x] Unit test: Module constructs, starts, stops without leaking threads.
- [x] `make check` is green; `make test-integration` flakiness is pre-existing on `main` (see caveat above).

## Out of scope

- Migrating any MCP tool methods onto the Module (#257-#260).
- The import-graph guard test (#262).
- `CorpusIntake` dependency anticipated for MVP2/MVP3.

### DIFF
--- a/src/lithos/cognitive_memory.py
+++ b/src/lithos/cognitive_memory.py
@@ -1,0 +1,132 @@
+"""CognitiveMemory — agent-facing surface of LCMA as a Module facade.
+
+See docs/adr/0005-cognitive-memory-module.md.
+
+This slice (issue #255) lands the seam and lifecycle ordering only.
+``CognitiveMemory`` owns the ``StatsStore`` and the ``EnrichWorker``;
+``LithosServer.initialize()`` constructs the Module after the projection
+is ready and ``LithosServer.shutdown()`` stops it first. Public read /
+write methods (retrieve, edge_*, reinforce_*, etc.) migrate in
+subsequent slices (#257-#260); existing MCP tool handlers continue to
+call LCMA internals directly through server-level aliases.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from lithos.lcma.enrich import EnrichWorker
+from lithos.lcma.stats import StatsStore
+
+if TYPE_CHECKING:
+    from lithos.config import LithosConfig
+    from lithos.coordination import CoordinationService
+    from lithos.events import EventBus
+    from lithos.graph import KnowledgeGraph
+    from lithos.knowledge import KnowledgeManager
+    from lithos.provenance import ProvenanceProjection
+    from lithos.search import SearchEngine
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["CognitiveMemory"]
+
+
+class CognitiveMemory:
+    """Module facade over LCMA's agent-facing surface (ADR-0005).
+
+    Owns the ``StatsStore`` lifecycle and the ``EnrichWorker``. Public
+    retrieve / edge / reinforcement methods are migrated in subsequent
+    slices; this slice exposes only the lifecycle.
+
+    Construction follows ADR-0002's eager-init pattern: prefer
+    :meth:`create` over calling ``__init__`` directly. ``coordination``
+    is required because ``EnrichWorker`` depends on it; ADR-0005 lists
+    six logical deps but the worker's operational dep is the seventh.
+    """
+
+    def __init__(
+        self,
+        config: LithosConfig,
+        knowledge: KnowledgeManager,
+        search: SearchEngine,
+        graph: KnowledgeGraph,
+        projection: ProvenanceProjection,
+        event_bus: EventBus,
+        coordination: CoordinationService,
+    ) -> None:
+        self._config = config
+        self._knowledge = knowledge
+        self._search = search
+        self._graph = graph
+        self._projection = projection
+        self._event_bus = event_bus
+        self._coordination = coordination
+
+        # Module-internal stores. ``StatsStore`` is constructed eagerly but
+        # opened in :meth:`start` per issue #255 ("start() opens the
+        # internal StatsStore and starts the EnrichWorker").
+        self._stats_store: StatsStore = StatsStore(config)
+        self._enrich_worker: EnrichWorker | None = None
+        self._started: bool = False
+
+    @classmethod
+    async def create(
+        cls,
+        config: LithosConfig,
+        knowledge: KnowledgeManager,
+        search: SearchEngine,
+        graph: KnowledgeGraph,
+        projection: ProvenanceProjection,
+        event_bus: EventBus,
+        coordination: CoordinationService,
+    ) -> CognitiveMemory:
+        """Construct the Module eagerly.
+
+        StatsStore open and EnrichWorker start are deferred to
+        :meth:`start`. Returning from ``create`` guarantees the Module
+        is fully wired but not yet running — callers must invoke
+        :meth:`start` before issuing work.
+        """
+        return cls(
+            config=config,
+            knowledge=knowledge,
+            search=search,
+            graph=graph,
+            projection=projection,
+            event_bus=event_bus,
+            coordination=coordination,
+        )
+
+    async def start(self) -> None:
+        """Open the StatsStore and start the EnrichWorker.
+
+        Raises ``RuntimeError`` if called twice without an intervening
+        :meth:`stop`. When ``config.lcma.enabled`` is ``False`` the
+        StatsStore is still opened (its receipts/working-memory tables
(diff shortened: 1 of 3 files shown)


davesnowdon:
## Integration suite re-verified — clean run matches baseline exactly

Re-ran `make test-integration` on this branch with `-v` to get verbose output. Result:

```
252 passed, 1139 deselected, 93 errors in 1126.98s (0:18:46)
```

This **exactly matches baseline `main`** (`252 passed, 93 errors in 1124.57s`, verified by stashing this branch's changes locally). All 93 errors are confined to:

- `tests/test_telemetry.py::TestTelemetryIntegration::*` — `torch.OutOfMemoryError: CUDA out of memory` (GPU pressure from the embedding model held by the previous test run, unrelated to this PR)
- `tests/test_smoke.py::*` — `TypeError: _NoOpSpan.set_status() takes 2 positional arguments but 3 were given` (pre-existing telemetry stub mismatch)

The earlier "5 failed, 168 errors" run I noted in the PR description was a transient flake where GPU OOM cascaded into setup errors for `test_server.py` fixtures further down the run. The current verbose run on the same code shows the cascade does not happen — the failures are stable on `test_telemetry.py` / `test_smoke.py` and don't propagate.

**Net: this branch does not change the integration test pass/fail count vs main.** Both pre-existing failure groups are tracked separately from #255.

davesnowdon:
## Addressed the review (commit 199ff10)

Thanks for the careful read — all three points are real and now fixed:

### 1. Public seam now matches ADR-0005 / #255 exactly (six args)

`CognitiveMemory.create` is back to the spec'd shape:
```python
await CognitiveMemory.create(config, knowledge, search, graph, projection, event_bus)
```

The `EnrichWorker` construction inside `start()` still needs a `CoordinationService` that ADR-0005 doesn't include in the Module's logical deps. A transitional `attach_coordination(coordination)` setter accepts it between `create` and `start`. When LCMA is enabled and coordination isn't attached, `start()` raises explicitly — loud, not silent. The setter is documented as the transitional bridge that disappears when coordination consolidates into the Module per ADR-0005's "Anticipated evolution" section. Subsequent slices (#257–#260) can build against the stable six-arg seam without re-deciding it.

### 2. `start()` now actually opens the StatsStore in the real path

Good catch — the issue's stated lifecycle invariant was being silently bypassed because `LithosServer.initialize()` called `self.stats_store.refresh_cached_counts()` (which hits `_ensure_open()`) before `await self.memory.start()`. I moved the LCMA gauge priming and `register_lcma_metrics` block to run **after** `memory.start()`, so the explicit `start()` is the first thing that touches the store. There's also a new pre-condition assert in `test_start_opens_stats_store_and_starts_worker` (`assert memory._stats_store._opened is False` before calling `start()`) to make the contract checkable in the unit test too.

### 3. Subprocess no-thread-leak test added

`TestProcessLifecycleValidation::test_create_start_stop_exits_cleanly_in_subprocess` mirrors `tests/test_server_lifecycle.py:10`. The subprocess runs three `create → attach_coordination → start → stop → projection.close` cycles and asserts the interpreter exits within 20s. A hung worker thread or unclosed aiosqlite handle fails on the exact symptom the acceptance criterion targets.

### Verification on this commit

- `make lint` — green
- `make typecheck` — green (pyright, 0 errors)
- `make test` — green, 1142 unit tests pass (1139 prior + 3 new on this module)
- `tests/test_cognitive_memory.py` — all 12 pass including the subprocess test
- `tests/test_server_lifecycle.py` — still green after the lifecycle reorder

Re-running `make test-integration` in the background; will follow up if it shows anything beyond the pre-existing telemetry/smoke baseline.

PR #301, opened 2026-05-25
fix(write): treat .md-ending paths as explicit filenames (closes #300)

### LINKED ISSUE #300: lithos_write: `path` silently treats `.md`-ending values as directories, producing nested layout
## Summary

`lithos_write`'s `path` parameter is documented as "Subdirectory path (e.g., `procedures`)" (`src/lithos/server.py:1037`) and implemented in `KnowledgeManager.create` as a directory under which `slugify(title) + ".md"` is always appended (`src/lithos/knowledge.py:1046-1047`). If a caller intuits the more common convention (full file path, ending in `.md`), Lithos silently treats the entire `.md`-suffixed string as a directory name, producing a doc nested one level deeper than intended.

## Reproduction

```bash
lithos_write \
  path="projects/slice-5-test-gammer/slice-5-test-gammer-project-context.md" \
  title="Slice 5 Test Gammer" \
  content="…" \
  agent="some-agent"
```

**Expected** (intuited by caller): doc lands at
```
knowledge/projects/slice-5-test-gammer/slice-5-test-gammer-project-context.md
```

**Actual**: doc lands at
```
knowledge/projects/slice-5-test-gammer/slice-5-test-gammer-project-context.md/slice-5-test-gammer.md
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ a directory whose name ends in .md
```

Reproduced on `agent-lore/lithos@07304b1` (current `main`).

## Root cause

`src/lithos/knowledge.py:1046-1047`:

```python
slug = slugify(title)
file_path = Path(path) / f"{slug}.md" if path else Path(f"{slug}.md")
```

The expression `Path(path) / f"{slug}.md"` unconditionally treats `path` as a directory, regardless of suffix. `_resolve_safe_path` (called immediately after) doesn't reject `.md`-suffixed directory components — they're valid filesystem names, just confusing ones.

There's no `invalid_input` branch in `KnowledgeManager.create` or in the `lithos_write` server-side validation that catches "looks like a file path, not a directory." So callers get success + a quietly wrong layout.

## Why it's load-bearing

Three behaviours combine to make this a sharp edge:

1. **Filename is title-derived, not caller-controlled.** A caller who wants a specific filename has no way to request it — the docs don't say so, and the failure mode (when they try) is silent acceptance with double-nesting.
2. **Wide intuition gap.** "Path" in most file-oriented APIs means full file path. Directory-only semantics are unusual without a parameter name like `dir`, `parent_dir`, or `subdirectory`.
3. **No round-trip warning.** The `lithos_read` response carries the actual stored path, so a careful caller would notice — but the bidirectional-sync code paths that have surfaced this don't read back immediately after writing every doc.

Surfaced in `lithos-loom`'s `project create` CLI (passes a full `.md`-ending path; got bitten during Slice 5 soak — `lithos-loom` issue forthcoming). Loom is fixing its own side regardless, but the underlying ambiguity will keep biting other callers.

### Option A (recommended) — accept full file paths when suffix is present

If `path` ends in `.md`, treat it as a complete relative file path. Otherwise, fall back to current directory-semantics. Two-line change:

```python
slug = slugify(title)
if path and path.endswith(".md"):
    file_path = Path(path)
else:
    file_path = Path(path) / f"{slug}.md" if path else Path(f"{slug}.md")
```

Backwards-compatible: existing callers passing bare directories see no change. Callers passing `.md`-ending paths get the layout they intuited. Update the `lithos_write` docstring to describe both modes.

### Option B — reject `.md`-ending `path` with `invalid_input`

Refuse the ambiguous shape rather than guess. Error message: "`path` is a subdirectory; the filename is derived from `title`. To control the filename, set the title accordingly." Cleanest semantically, but breaks callers who've already adapted to the current behavior (if any) and gives no escape hatch for callers who genuinely want a specific leaf name.

### Option C — add a `filename` parameter, deprecate `.md`-ending `path`

```python
async def lithos_write(
    *,
    title: str,
    content: str,
    agent: str,
    path: str | None = None,       # subdirectory (existing semantics, preserved)
    filename: str | None = None,   # NEW: overrides slugify(title) + ".md"
    ...
)
```

Most explicit; biggest surface change. Probably overkill if Option A covers the cases that matter.

## Acceptance

Whichever option is taken, the docstring at `server.py:1037` should be updated to describe the resolved semantics unambiguously, and a test should exist that exercises a `.md`-ending `path` input asserting either:

- (Option A) the doc lands at the supplied path exactly, OR
- (Option B) `lithos_write` returns `invalid_input`, OR
- (Option C) the new `filename` param overrides the slugified default.

## Severity

Medium. Doesn't lose data — the doc IS persisted, just in a surprising place — but the silent layout drift means consumers' canonical-path assumptions can quietly diverge from reality, which is the kind of thing that compounds over time in a KB.

davesnowdon:
## Summary

- `lithos_write` now accepts a full relative file path ending in `.md` and stores the doc at exactly that path; bare-directory inputs behave unchanged.
- Any path with a non-final segment ending in `.md` is rejected with `invalid_input` — making it structurally impossible to create directories whose names end in `.md`.
- `lithos_write` `path` docstring updated to describe both modes.

Closes #300.

## Why

Callers intuiting the more common "full file path" convention silently got a doc nested one level too deep, under a directory whose name ended in `.md`. The maintainer has had to manually clean up such directories from earlier callers, so this is shaping the on-disk knowledge base.

Implements Option A from the issue, with the additional invariant — requested in follow-up — that no directory ending in `.md` may be created via `lithos_write` going forward.

## Behavioural matrix

| `path` input | Before | After |
|---|---|---|
| `None` / `""` | `slugify(title).md` at root | unchanged |
| `"procedures"` | `procedures/slugify(title).md` | unchanged |
| `"a/b/c"` | `a/b/c/slugify(title).md` | unchanged |
| `"procedures/my-doc.md"` | `procedures/my-doc.md/slugify(title).md` ⚠️ | `procedures/my-doc.md` ✅ |
| `"root-level.md"` | `root-level.md/slugify(title).md` ⚠️ | `root-level.md` ✅ |
| `"projects/foo.md/bar"` | `projects/foo.md/bar/slugify(title).md` ⚠️ | `invalid_input` ✅ |
| `"a.md/b.md"` | `a.md/b.md/slugify(title).md` ⚠️ | `invalid_input` ✅ |
| `"../outside"` / `"../outside.md"` | `ValueError` (traversal) | unchanged |

Clients that operate correctly today see no behaviour change.

## Test plan

- [x] `make test` (unit) — 1202 passed
- [x] `make test-integration` — 363 passed
- [x] `make lint` — clean
- [x] `make typecheck` — clean
- [x] `make check` — all green
- [x] Backwards-compat anchors added: `None`, `procedures`, `a/b/c`
- [x] New `.md` happy path: nested + root
- [x] New rejection paths: intermediate `.md`, multiple `.md`
- [x] Traversal interaction: `../outside.md` still raises `ValueError`
- [x] Integration: `.md`-ending path round-trips through intake; bad path → `invalid_input` at MCP tier

### DIFF
--- a/docs/SPECIFICATION.md
+++ b/docs/SPECIFICATION.md
@@ -400,7 +400,7 @@ Parameters are flat at the MCP boundary but grouped below by role to aid discove
 | `id` | string | No | UUID to update existing; omit to create new |
 | `tags` | string[] | No | List of tags |
 | `confidence` | float | No | Confidence score 0-1 (default: 1.0) |
-| `path` | string | No | Subdirectory path (e.g., "procedures") |
+| `path` | string | No | Either a subdirectory (e.g. `"procedures"`) under which the filename is derived from `title` (slugified) and `.md` appended, OR a full relative file path ending in `.md` (e.g. `"procedures/my-doc.md"`) used verbatim as the filename. Intermediate path segments may not end in `.md` — such inputs return `status="invalid_input"`. Paths that resolve to an already-owned file return `status="path_collision"`. |
 
 *Provenance:*
 
@@ -442,13 +442,15 @@ The error code is the canonical top-level `status` value (e.g. `status="slug_col
 
 `{ status: "updated", id: string, path: string, version: int, warnings: string[] }`
 
-`{ status: "duplicate", duplicate_of: { id, title, source_url }, message: string, warnings: string[] }`
+`{ status: "duplicate", duplicate_of: { id, title, source_url }, message: string, warnings: string[] }`  *(source-URL dedup only — never used for filesystem-path conflicts; see `path_collision` below)*
 
 `{ status: "invalid_input", message: string, warnings: string[] }`
 
 `{ status: "content_too_large", message: string, warnings: string[] }`
 
 `{ status: "slug_collision", message: string, existing_id: string, warnings: string[] }`
+
+`{ status: "path_collision", message: string, existing_id: string, warnings: string[] }`  *(another doc already owns the requested file path — e.g. caller passed an explicit `path` ending in `.md` that's already in use)*
 
 `{ status: "version_conflict", message: string, current_version: int, warnings: string[] }`
 
--- a/src/lithos/intake.py
+++ b/src/lithos/intake.py
@@ -154,12 +154,14 @@ class WriteOutcome:
         "version_conflict",
         "content_too_large",
         "slug_collision",
+        "path_collision",
         "error",
     ]
     document: KnowledgeDocument | None = None
     duplicate_of: DuplicateInfo | None = None
     current_version: int | None = None
     slug_collision_existing_id: str | None = None
+    path_collision_existing_id: str | None = None
     message: str | None = None
     warnings: list[str] = field(default_factory=list)
 
@@ -366,6 +368,7 @@ class CorpusIntake:
                     document=result.document,
                     duplicate_of=result.duplicate_of,
                     current_version=result.current_version,
+                    path_collision_existing_id=result.path_collision_existing_id,
                     message=result.message,
                     warnings=list(result.warnings),
                 )
--- a/src/lithos/knowledge.py
+++ b/src/lithos/knowledge.py
@@ -447,11 +447,16 @@ class WriteResult:
     """Structured result type for create/update operations.
 
     Error outcomes use the error code as the canonical ``status`` value
-    (``invalid_input`` / ``version_conflict`` / ``content_too_large``)
-    rather than ``status="error"`` plus a separate discriminator field.
-    The plain ``"error"`` status remains as a generic fallback for
-    unforeseen failures and ``slug_collision`` is raised as a
+    (``invalid_input`` / ``version_conflict`` / ``content_too_large`` /
+    ``path_collision``) rather than ``status="error"`` plus a separate
+    discriminator field. The plain ``"error"`` status remains as a generic
+    fallback for unforeseen failures and ``slug_collision`` is raised as a
     ``SlugCollisionError`` exception (not represented in WriteResult).
+
+    Per ``docs/plans/unified-write-contract.md``: ``"duplicate"`` is specific
+    to source URL dedup; filesystem-level conflicts use ``"path_collision"``
+    and carry the existing doc's id in ``path_collision_existing_id``,
+    mirroring ``slug_collision_existing_id`` on the intake-layer outcome.
     """
 
     status: Literal[
@@ -462,12 +467,14 @@ class WriteResult:
         "invalid_input",
         "version_conflict",
         "content_too_large",
+        "path_collision",
     ]
     document: KnowledgeDocument | None = None
     warnings: list[str] = field(default_factory=list)
     message: str | None = None
     duplicate_of: DuplicateInfo | None = None
     current_version: int | None = None
+    path_collision_existing_id: str | None = None
 
 
 def slugify(text: str) -> str:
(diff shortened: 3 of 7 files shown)


davesnowdon:
Follow-up commit ([`4307a99`](https://github.com/agent-lore/lithos/commit/4307a99)) addressing review feedback flagging a High-severity regression:

In the new explicit-`.md` filename mode, two creates with different titles could resolve to the same `file_path` (e.g. both passing `path="same.md"`). The existing `_slug_to_id` check keys on `slugify(title)` and didn't fire for distinct titles, so the second write silently overwrote the first file and corrupted the path/id indexes.

**Fix:** added a `_path_to_id` collision check immediately after the slug check in `KnowledgeManager.create`. On hit, returns `WriteResult(status="duplicate", duplicate_of=..., message=...)`, reusing the existing URL-duplicate envelope that intake + server already understand. Slug-check stays first to preserve `slug_collision` semantics for the title-duplicate case.

**New tests:**
- Unit: duplicate explicit-`.md` create → `status="duplicate"`, original file & both `read(id)` / `read(path)` still resolve to the first doc.
- Unit: mixed collision — `path="procedures"` + title `"Foo"` → `procedures/foo.md`, then `path="procedures/foo.md"` from a different title → blocked.
- Unit: round-trip after a fresh `KnowledgeManager` reloads from disk; collision guard still fires post-restart (proves the disk scan correctly seeds `_path_to_id` for explicit filenames).
- Integration: collision through full MCP `lithos_write` returns `status="duplicate"` with `duplicate_of.{id,title}` populated; `lithos_read(id=first)` still returns the first body.

Locally: `make lint` ✅ · `make typecheck` ✅ · `make test` ✅ (1205 passed) · `make test-integration` ✅ (364 passed).

davesnowdon:
Second follow-up ([\`16a8697\`](https://github.com/agent-lore/lithos/commit/16a8697)) addressing review feedback on contract drift.

The previous fix returned \`status="duplicate"\` for path conflicts, but [\`docs/plans/unified-write-contract.md:134\`](../blob/main/docs/plans/unified-write-contract.md#L134) reserves \`duplicate\` for source-URL dedup and lists [\`path_collision\`](../blob/main/docs/plans/unified-write-contract.md#L151) as the machine-readable code for filesystem conflicts. Reusing \`duplicate\` would have leaked a nullable \`source_url\` into the envelope of a non-URL response.

**This commit:**
- Adds \`"path_collision"\` to \`WriteResult.status\` and \`WriteOutcome.status\` Literals.
- Adds \`path_collision_existing_id: str | None\` to both, mirroring the existing \`slug_collision_existing_id\`.
- \`KnowledgeManager.create\` now returns \`WriteResult(status="path_collision", path_collision_existing_id=..., message=...)\` — no more synthetic \`duplicate_of\`.
- \`CorpusIntake.write\` passthrough forwards the new field.
- \`lithos_write\` MCP handler adds a \`path_collision\` branch mirroring \`slug_collision\`'s envelope: \`{status, message, existing_id, warnings}\`. No \`duplicate_of\` leakage.

**New / updated tests:**
- Unit + integration tests for path-collision now assert \`status="path_collision"\`, \`path_collision_existing_id == first_id\`, and \`duplicate_of is None\`.
- New conformance test \`test_write_path_collision_envelope_contract\` in \`tests/test_integration_conformance.py\` pins the exact public envelope key set and that \`lithos_read(id=first)\` still returns the first body.

**Locally:** \`make lint\` ✅ · \`make typecheck\` ✅ · \`make test\` ✅ (1205 passed) · \`make test-integration\` ✅ (365 passed, one more than the previous run thanks to the new conformance test).

davesnowdon:
Docs follow-up ([\`25402b5\`](https://github.com/agent-lore/lithos/commit/25402b5)) addressing the remaining review note on \`docs/SPECIFICATION.md\`:

- Added the \`path_collision\` envelope alongside \`slug_collision\` in the \`lithos_write\` Returns section.
- Added a clarifying note next to the \`duplicate\` envelope that it is source-URL-only — never used for filesystem-path conflicts.
- Replaced the terse \`path\` argument description with the full dual-mode semantics (directory mode vs explicit-\`.md\` mode) and the two rejection cases (\`invalid_input\` for intermediate \`.md\` segments, \`path_collision\` for already-owned paths).

\`unified-write-contract.md\` was already consistent (lists \`path_collision\` as a Core code, reserves \`duplicate\` for URL dedup), so no change there.

Docs-only change — no code or test diff.

PR #341, opened 2026-06-13
docs(task-graph): raise the cancelled-blocker readiness semantic

davesnowdon:
## Summary

A follow-up edit to the [task-graph coordination extension](docs/plans/task-graph-coordination-extension.md) proposal (#339/#340), raising one semantic gap found while planning a consumer (Lithos Loom) against it.

## The gap

§5.2 defines a `blocks` edge as *"`to_task_id` cannot be considered ready **while `from_task_id` is open**"*. That resolves the edge the instant the predecessor leaves `open` — **including when it ends `cancelled`**. So `lithos_task_ready` would return a dependent whose predecessor was *abandoned*, not completed.

For a strict-sequential chain — e.g. a decomposed PRD where story N+1 builds on story N — that's a correctness bug: cancelling story N would spuriously ready story N+1. It also silently regresses the pre-extension convention every current consumer uses, which gates on `status == "completed"` and *fails* downstream work when a blocker is cancelled (the `[BlockerFailed]` pattern).

## The fix (semantics clarification, not new machinery)

Folded consistently into the affected sections:

- **§5.2** — a `blocks` edge is satisfied only when the predecessor is `completed`; a `cancelled` predecessor does not satisfy it.
- **§8** — readiness predicate becomes "every blocking predecessor `completed`" (not "not open"); new **Blocker-failure policy**: a `cancelled` predecessor leaves dependents permanently blocked, excluded from `ready`, edge retained — the same treatment backfill cycles already get. Lithos stays **passive**: it refuses to call them ready and explains why; the orchestrator decides whether to re-open, re-route, or cancel the subtree.
- **§6.1** — adds a `blocker_unsatisfiable` blocker kind and a `status` field so clients can tell "not yet" (predecessor `open`) from "needs intervention" (predecessor `cancelled`).
- **§12.4** — new risk entry documenting the regression and mitigation.
- **§13** — a test bullet for the cancelled-blocker case.

Net change: the readiness predicate moves from "predecessor not open" to "predecessor completed", plus one new blocker kind. No schema change.

## Context

Surfaced while drafting the post-extension orchestration plan for Lithos Loom (agent-lore/lithos-loom#97), which relies on `lithos_task_ready` for dispatch gating and therefore can't tolerate a cancelled blocker reading as ready.

### DIFF
--- a/docs/plans/task-graph-coordination-extension.md
+++ b/docs/plans/task-graph-coordination-extension.md
@@ -151,7 +151,7 @@ MVP edge types. An edge type is only accepted on write once the phase that imple
 Accepted from Phase 1:
 
 - `blocks`
-  Meaning: `to_task_id` cannot be considered ready while `from_task_id` is open. Blocking.
+  Meaning: `to_task_id` cannot be considered ready until `from_task_id` reaches `completed`. A blocker that ends in a **non-`completed` terminal state** (`cancelled`) does **not** satisfy the edge — `to_task_id` stays blocked (surfaced via `lithos_task_blocked` as `blocker_unsatisfiable`; see §6.1, §8), because the work the dependent was waiting on never happened. Resolving readiness on "predecessor is merely not `open`" would instead let a cancelled blocker spuriously ready its dependents — a correctness bug for strict-sequential chains (see §12.4). Blocking.
 - `parent_child`
   Meaning: `from_task_id` is the parent; `to_task_id` is the child. Purely structural — it never blocks the child. Epic roll-up rules (e.g. a parent cannot close while children are open) operate in the reverse direction and are deferred to Phase 4.
 - `discovered_from`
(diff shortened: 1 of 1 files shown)


davesnowdon:
Thanks — both addressed in 3086759:

- **Blocker schema inconsistency (§6.1, L293)** → a `blocker_unsatisfiable` blocker now carries the cancelled predecessor in `task_id` (with its `status`), not in free-text `message`. `message` stays human-readable. (`cycle` blockers keep using `message` since they carry a *set* of ids.)
- **Prose still describing the old "open predecessor" readiness predicate** → updated the three lagging spots to the clarified semantic: `lithos_task_ready` ("no unsatisfied blocking edges — every predecessor `completed`"), the `lithos_task_blocked` behaviour bullet ("unsatisfied predecessor — not yet `completed`"), and the `lithos_task_spawn` `blocks` relation ("blocked until the source is `completed`; a cancelled source → `blocker_unsatisfiable`").

PR #384, opened 2026-07-12
fix: route enrich/quarantine/supersede note writes through Corpus intake (PR1a, task 681ac952)

davesnowdon:
## Summary

**PR1a of task `681ac952`** (architecture-review epic `f80b6eb4`). Fixes a real
Drift bug: three note writes bypassed `CorpusIntake` and mutated the Corpus via
`KnowledgeManager.update` directly, so the Search index and link graph were never
re-synced and no `NOTE_UPDATED` event fired — the derived views stayed stale until
the next Reconcile.

The three bypassing writes, now routed through `CorpusIntake.note_update`:

| Site | Write |
|------|-------|
| `lcma/enrich.py` `_extract_entities` | entities / entities_extractor |
| `cognitive_memory.py` `reinforce_misleading` | `lcma_status="quarantined"` |
| `cognitive_memory.py` `conflict_resolve` | `supersedes` (on the winner) |

## What changed

- **`NoteUpdateRequest`** gains `supersedes` / `entities` / `entities_extractor`
  (all `_UNSET`-defaulted) so the frontmatter-only patch seam can carry the
  LCMA-internal fields; `note_update` forwards them to `KnowledgeManager.update`.
- **`EnrichWorker`** takes a `CorpusIntake` dependency (wired in
  `CognitiveMemory.start`) and routes its entity write through it.
- **Loop-break.** Those writes now emit events that `EnrichWorker` subscribes to,
  which would re-enqueue the just-written node (enrich→write→event→enrich). New
  `ENRICH_AGENT = "lithos-enrich"` sentinel stamps enrichment-originated writes,
  and `_handle_event` drops any event carrying that actor. This makes the
  loop-break **explicit**, replacing the incidental protection that previously
  relied only on the entities-extractor marker's idempotency.

## Architecture budget

Routing enrich through intake introduces the **LCMA → Intake** edge, so
`cross_component_edges` goes 63 → 64 and Intake joins the Core SCC. This is
intrinsic to the fix; the budget is raised in `docs/architecture.toml` with a
rationale and the generated diagrams regenerated (guardrails green).

## Scope & follow-ups

- **Deferred to PR1b** (same task, after this merges): route provenance edges
  through `ProvenanceProjection` plan/apply, dedupe the two `related_to` policies,
  and the edge_store plumbing cleanup (public `edge_store` property, retire
  reach-throughs, unify the two `EDGE_UPSERTED` payloads).
- **Out of scope:** `cli.py` `extract-entities` still writes directly — the CLI
  can't cross the lcma import seam and needs its own intake; it folds into Task 6's
  CLI pipeline factory.

## Test plan

- [x] `make check` — typecheck (0 errors), unit tests (1550 passed), lint (clean)
- [x] `make test-integration` — 428 passed
- [x] `make diagrams` — 48 guardrail tests pass; regenerated docs committed
- New regression tests:
  - `test_enrich_worker.py::TestExtractEntities::test_extract_entities_reindexes_search_via_intake` — entity write now re-indexes Search + syncs graph (Drift closed) while still persisting frontmatter.
  - `test_enrich_worker.py::TestSelfOriginatedEventFilter` — `ENRICH_AGENT` events are dropped; genuine agent writes still enqueue.
  - `test_cognitive_memory.py::...::test_quarantine_reindexes_search_via_intake` — quarantine re-indexes Search.
  - `test_cognitive_memory.py::...::test_superseded_records_link_and_reindexes_via_intake` — first positive superseded test: records the link on the winner **and** re-indexes.
  - `test_intake.py::test_note_update_forwards_lcma_fields_and_reindexes` / `test_note_update_forwards_supersedes` — new fields round-trip, body untouched.

### DIFF
--- a/docs/architecture.toml
+++ b/docs/architecture.toml
@@ -85,8 +85,8 @@ Foundation  = ["Events", "Config", "Telemetry", "Errors", "Logging"]
 # is deliberately NOT pinned to the exact current high-water mark. Ratchet the
 # *count* of oversized modules down via modules_over_800_lines instead.
 [budgets]
-cross_component_edges  = 63    # measured 2026-07; direction: down
-component_cycles       = 1    # one 6-component SCC in Core (Coordination/Graph/Knowledge/LCMA/Provenance/Search); goal: shrink it
+cross_component_edges  = 64    # measured 2026-07; +1 (LCMA→Intake, task 681ac952: enrich/cognitive-memory note writes now route through CorpusIntake to fix Drift); direction: down
+component_cycles       = 1    # one 7-component SCC in Core (Coordination/Graph/Intake/Knowledge/LCMA/Provenance/Search); Intake joined via LCMA→Intake (681ac952); goal: shrink it
 module_cycles          = 2    # knowledge<->graph<->search<->provenance<->lcma.edges, server<->tools; goal: 0
 modules_over_800_lines = 11   # incl. tools.tasks (985) split from server in #374; direction: down
 max_module_lines       = 2800 # stop-loss; largest today is coordination.py at 2675
--- a/docs/generated/architecture.md
+++ b/docs/generated/architecture.md
@@ -89,6 +89,7 @@ graph TD
   LCMA --> Errors
   LCMA --> Events
   LCMA --> Graph
+  LCMA --> Intake
   LCMA --> Knowledge
   LCMA --> Provenance
   LCMA --> Search
@@ -127,10 +128,10 @@ graph TD
   linkStyle 45 stroke:#bbb
   linkStyle 47 stroke:#bbb
   linkStyle 48 stroke:#bbb
-  linkStyle 53 stroke:#bbb
   linkStyle 54 stroke:#bbb
-  linkStyle 58 stroke:#bbb
+  linkStyle 55 stroke:#bbb
   linkStyle 59 stroke:#bbb
-  linkStyle 61 stroke:#bbb
+  linkStyle 60 stroke:#bbb
   linkStyle 62 stroke:#bbb
+  linkStyle 63 stroke:#bbb
 ```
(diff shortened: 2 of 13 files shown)


davesnowdon:
Thanks for the thorough review. Addressed in 7c2fab5:

**Finding 2 (stale docstring)** — Fixed: `reinforce_misleading` now names `CorpusIntake.note_update`.

**Finding 1 (ENRICH_AGENT spoofing)** — Agreed it's real. The loop-break overloads the caller-facing `agent` string, so an external write with `agent="lithos-enrich"` would also be skipped — its node is deferred to the periodic `full_sweep`, not lost, but that's a silent gap. The robust fix is a **non-spoofable event-origin marker set by the intake** (callers cannot forge event payloads) rather than the agent string. I've deferred that to **PR1b**, which is already reworking the `EDGE_UPSERTED` payloads — landing the origin marker there does it once, cleanly, instead of a half-version here. For PR1a I've documented the reservation caveat inline at `ENRICH_AGENT`. Tracked on task `681ac952`.

**Test gap (connect the halves)** — Added `test_extract_entities_emits_self_event_that_worker_drops`: `_extract_entities` emits a `NOTE_UPDATED` stamped `ENRICH_AGENT`, and feeding that exact emitted event back through `_handle_event` is dropped — closing `enrich→write→event→enrich` end to end. Also strengthened the intake `note_update` test to assert graph sync + event `agent`.

**Q: Is `lithos-enrich` a globally reserved agent ID?** Yes — system-reserved, mirroring `WATCHER_AGENT = "watcher"`. Not enforced at the write boundary today; PR1b's intake-set origin marker removes the need for that enforcement (the loop-break stops depending on the caller-facing string).

**Q: Quarantine dropped vs supersede re-enqueued — intentional?** Yes. Quarantine is a system status flip attributed to `ENRICH_AGENT` → dropped (re-enriching a node from its own just-applied quarantine is pointless; `full_sweep` still covers it). Supersede is a genuine agent action attributed to the real `resolver` → re-enqueued, because the winner note legitimately changed and should be enriched.

`make check` (unit 1550 + guardrails) and `make test-integration` (428) remain green.

PR #385, opened 2026-07-12
refactor: consolidate edge-write policy + non-spoofable enrich loop-break (PR1b, task 681ac952)

davesnowdon:
## Summary

**PR1b of task `681ac952`** — the clean, no-behaviour-change half of Task 2 (edge-policy consolidation). PR1a fixed the Drift bug + added the loop-break; this PR consolidates the edge-write surfaces. The behaviour-changing provenance-projection migration is deferred to **PR1c**.

Three consolidations, all low-risk:

### 1. Loop-break hardening (PR1a review Finding 1)

The `enrich→write→event→enrich` loop-break now keys on a new **non-spoofable `LithosEvent.origin` marker** set by the intake, not the caller-facing `agent` string. Previously an external write passing `agent="lithos-enrich"` would have been dropped as a self-event; now it isn't. `origin` is a code-level seam parameter on `CorpusIntake.note_update` — absent from the MCP tool schema **and** the SSE wire, so callers cannot forge it. `ENRICH_AGENT` remains as contributor attribution only.

### 2. `related_to` policy dedup

`EnrichWorker._consolidate_task` and `CognitiveMemory.reinforce_between` shared a create-or-strengthen skeleton (differing only in initial weight / `provenance_type`). Extracted to `lcma/edge_reinforce.reinforce_related_edge`, which returns `(edge_id, created)` so each caller keeps its own create-vs-strengthen logging.

### 3. 11c edge_store plumbing

- Added the public **`ProvenanceProjection.edge_store` property** (promised in ADR-0005) and retired the **6 `_projection._edge_store` reach-throughs** in `cognitive_memory.py`.
- Unified the two divergent `EDGE_UPSERTED` payloads behind **`events.make_edge_upserted_event`** — one shape (`edge_id`, `from_id`, `to_id`, `type`, `namespace`, `conflict_state`). Previously `assert_edge` had `namespace`/`agent` but no `conflict_state`, and `conflict_resolve` the reverse.

## Not in this PR

- **PR1c** (behaviour change): route enrich's `derived_from` writes through `ProvenanceProjection` plan/apply (needs a new per-node projection method + injecting the projection into `EnrichWorker` + the asserted-key/drift handling the simpler helper lacks).
- `cli.py` extract-entities still writes directly (crosses no lcma seam; folds into Task 6's CLI factory).

## Guardrails

No new cross-component edge — `CognitiveMemory→LCMA` already existed, so importing `lcma.edge_reinforce` adds none; the new module maps to the LCMA component by prefix. `cross_component_edges` stays 64. Regenerated `docs/generated/` committed.

## Test plan

- [x] `make check` — typecheck 0 errors, unit **1556 passed**, lint clean
- [x] `make test-integration` — **428 passed**
- [x] `make diagrams` — 48 guardrails pass; regenerated docs committed
- New/updated tests:
  - `test_enrich_worker.py`: loop-break now keyed on `origin`; **`test_agent_string_alone_does_not_suppress`** proves the Finding-1 fix (agent string no longer suppresses); connecting test asserts the emitted event carries `origin`; `test_reinforce_related_edge_creates_then_strengthens` covers the shared helper.
  - `test_intake.py`: `test_note_update_stamps_internal_origin_on_event` (origin passthrough + default `""`); assert_edge emit test now asserts the unified `conflict_state` key.
  - `test_cognitive_memory.py`: `test_emits_unified_edge_upserted_payload` (conflict_resolve now carries agent/namespace/conflict_state); `test_projection_edge_store_property_exposes_store`.

### DIFF
--- a/docs/generated/components/Events.md
+++ b/docs/generated/components/Events.md
@@ -11,12 +11,13 @@ Internal event bus decoupling producers from consumers.
 
 | Module | Size | Classes | Functions |
 |---|---|---:|---:|
-| `lithos.events` | S | 2 | 0 |
+| `lithos.events` | M | 2 | 1 |
 
 ## Public API
 
 ### `lithos.events`
 - class `LithosEvent` — A typed event emitted by the Lithos event bus.
+- def `make_edge_upserted_event` — Build the canonical :data:`EDGE_UPSERTED` event.
 - class `EventBus` — In-memory event bus with filtered subscriptions and ring buffer history.
 
 ## Dependencies
--- a/docs/generated/components/LCMA.md
+++ b/docs/generated/components/LCMA.md
@@ -12,6 +12,7 @@ Lithos Cognitive Memory Architecture internals — scouts, retrieval, enrichment
 | Module | Size | Classes | Functions |
 |---|---|---:|---:|
 | `lithos.lcma` | XS | 0 | 0 |
+| `lithos.lcma.edge_reinforce` | XS | 0 | 1 |
 | `lithos.lcma.edges` | S | 0 | 0 |
 | `lithos.lcma.enrich` | L | 1 | 0 |
 | `lithos.lcma.entities` | M | 0 | 1 |
@@ -22,6 +23,9 @@ Lithos Cognitive Memory Architecture internals — scouts, retrieval, enrichment
 | `lithos.lcma.utils` | S | 1 | 1 |
 
 ## Public API
+
+### `lithos.lcma.edge_reinforce`
+- def `reinforce_related_edge` — Strengthen an existing ``related_to`` edge or create one.
 
 ### `lithos.lcma.enrich`
 - class `EnrichWorker` — In-process background worker that consumes events and drains enrichment work.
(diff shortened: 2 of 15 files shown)


davesnowdon:
Thanks — both addressed in a3cacd9:

**Finding 1 (positional compatibility)** — Agreed; appending is the right call. Moved `LithosEvent.origin` to the **last** dataclass field, so `type`/`agent`/`payload`/`tags`/`id`/`timestamp` keep their positions — `LithosEvent("note.updated", "agent", {...})` still binds the dict to `payload`. All internal callers set `origin=` by keyword, so nothing else changes. (You're right that I'd originally placed it after `agent` for readability; not worth the compatibility trap for a type that's in the generated public API.)

**Q: internal-only, or preserve compatibility?** Preserve compatibility — done by appending, per above.

**Nice-to-have (SSE contract test)** — Added `test_origin_not_emitted_on_wire` in `TestFormatSSE`: asserts `_format_sse` output contains neither the `origin` key nor the marker value, and that the parsed `data` JSON has no `origin`. Locks the "internal, never on the wire" contract against regressions.

Re-ran the touched suites (223 passed), guardrails (48), lint, and typecheck — all green.

PR #386, opened 2026-07-12
refactor: consolidate derived_from projection into ProvenanceProjection (PR1c, task 681ac952)

davesnowdon:
## Summary

**PR1c of task `681ac952`** — the final, behaviour-changing piece of Task 2, and the one I deliberately split out. The enrich worker projected `derived_from` edges through the raw `lcma/edges.py` helpers, duplicating the canonical-row logic `ProvenanceProjection` already owns and bypassing its asserted-key-collision blocking + column-drift resync. This routes enrich through the projection, making it the single owner.

### What changed

- **`ProvenanceProjection` gains public entry points:** `reconcile_corpus(docs)` and `reconcile_node(node_id, *, sources, namespace)`, over a new shared `_plan_actions` core. `reconcile_node` is the **per-node analogue** the projection lacked (the corpus-wide planner reads all `derived_from` edges; the node planner scopes to `from_id == node_id`). Callers pass pre-resolved `sources`/`namespace`, so the projection stays KnowledgeManager-decoupled.
- **`EnrichWorker` now takes the projection, not a raw `EdgeStore`:** `_enrich_node` → `reconcile_node` (resolving sources from KM's in-memory maps — no disk read), `full_sweep` → `reconcile_corpus` (loading the corpus once, shared with the entity heal), consolidation reaches the store via `projection.edge_store`.
- **Deleted `lcma/edges.py`** (the two helpers + a vestigial `EdgeStore` re-export); repointed `test_edges.py` to `lithos.edge_store`.

### Behaviour change (more correct)

The per-node path now matches the corpus-wide reconcile's semantics:
- **Asserted-key collisions are blocked** — an agent-asserted `derived_from` edge sharing the natural key survives. The old helper's unconditional upsert would have clobbered its provenance (dropping the actor). Covered by `TestReconcileNode.test_asserted_edge_at_same_key_is_not_clobbered`.
- **Only frontmatter-owned drifted rows are resynced** (not every desired row re-written), and a deleted node removes only its frontmatter-owned edges.

### Architecture win

Removing the `Provenance→LCMA` import **shrank the Core component cycle from 7 components to 4** — `Coordination`, `Intake`, and `LCMA` left the SCC (it's now just `Graph ↔ Knowledge ↔ Provenance ↔ Search`), reversing the expansion PR1a introduced. `cross_component_edges` dropped 64 → 63. Both budgets ratcheted **down** in `docs/architecture.toml`.

## Test plan

- [x] `make check` — typecheck 0 errors, unit **1558 passed**, lint clean
- [x] `make test-integration` — (running; PR1a/1b were 428 passed)
- [x] `make diagrams` — 48 guardrails pass (incl. tightened budgets); regenerated docs committed
- Tests: rewrote the former `_project_node_provenance` direct tests as `TestReconcileNode` (deleted-node / sync / no-op / **asserted-key-not-clobbered**) and `TestReconcileNodeResync` (frontmatter drift → resync); `EnrichWorker` fixtures now inject a `projection` sharing the test `EdgeStore`.

## Task 2 complete

With this, `681ac952` is done: **PR1a** (#384, Drift fix + loop-break), **PR1b** (#385, edge-policy consolidation + non-spoofable origin marker), **PR1c** (this, projection migration). Only out-of-scope follow-up: `cli.py` extract-entities still writes directly (folds into Task 6's CLI factory).


davesnowdon:
Thanks — Finding 1 is a good catch, and chasing it down showed the problem is **bigger than provenance and pre-exists this PR**. Both findings are addressed in `bd7f36e`.

## Finding 1 — fixed at the source, not guarded at the caller

You're right that the sweep's provenance source changed: `iter_doc_sources()` (in-memory map, always complete) → `list_all()` docs, and `list_all` really does swallow per-doc read failures — a bare `except Exception: pass`, not even logged.

But the exposure isn't the sweep's. `KnowledgeManager._scan_corpus()` — the canonical corpus reader behind **Reconcile** — used the *identical* two-call `list_all(limit=0)` → `list_all(limit=total)` idiom, and it feeds all three derived views. So on a partial read, `main` already:

- rebuilds **Tantivy + Chroma** from the short snapshot (`search.py:1881`, `doc_set_mismatch` → `full_rebuild`), silently dropping the unreadable docs from search;
- removes their **graph** nodes;
- deletes their **`derived_from`** edges.

Provenance was just the instance you happened to surface. A `len(docs) != total` guard in `full_sweep` would have fixed the symptom you saw and left the larger hole open — and my `full_sweep` was *also* hand-duplicating `_scan_corpus`'s body, which is what let the two drift apart in the first place.

So, to your first question — **provenance should require a complete corpus snapshot** — and so should search and graph:

- `CorpusScanError` (carries `expected`/`read`).
- `_scan_corpus` → public, strict **`scan_corpus()`**: raises when the read is short of what the metadata cache says exists.
- `plan_reconcile` propagates it. No plan is safer than a plan built from a partial snapshot.
- `full_sweep` now calls `scan_corpus()` (retiring the duplicated scan) and skips the provenance reconcile + entity heal on a partial read. Decay and WM eviction are already applied and unaffected.
- `list_all` still skips an unreadable note — one bad file shouldn't break a listing — but now **logs the cause** rather than swallowing it.

Note this changes `lithos reconcile` semantics: it now fails loudly on an unreadable note instead of quietly reindexing without it. That's the intended direction for a destructive operation, and it was an explicit call rather than a side effect.

Regression test added, and **verified to fail against the pre-fix `full_sweep`** (edge deleted) rather than merely passing after.

## Finding 2 — fixed

Correct, and worse than described: `mock_knowledge` is a bare `MagicMock`, so `get_cached_meta`/`get_doc_sources` returned MagicMocks and `_enrich_node` could hand *those* to `reconcile_node` with no test objecting. `TestEnrichNodeProvenanceHandoff` now drives the worker against a real `KnowledgeManager`:

- an existing node projects an edge carrying its **real source and namespace** (asserts `namespace == "research"`, so a default-namespace regression fails);
- an absent node resolves to empty sources and has its frontmatter-owned edges removed.

## Q2 — legacy `provenance_type IS NULL` rows: no migration needed

I checked rather than assumed. The deleted helper wrote `provenance_type="frontmatter"` **explicitly at both call sites** (`_project_node_provenance` and `_project_provenance_to_edges`), so there is no legacy NULL population of projection-owned rows to canonicalise.

NULL `derived_from` rows only arise from `lithos_edge_upsert`, whose `provenance_type` defaults to `None` (`tools/memory_edges.py:35`) — i.e. they're **agent assertions**. Treating those as asserted-forever is the intended ADR-0004/0006 contract (asserted rows are authoritative; the projection must not clobber them), not an oversight. That's the behaviour change `test_asserted_edge_at_same_key_is_not_clobbered` pins — and it's a fix: the old helper's unconditional upsert *would* have clobbered such a row.

## Nice-to-have — skipped

`Sequence[str]` over `list[str]` in `reconcile_node`: left as-is. Both call sites pass a concrete list, and the codebase's other reconcile entry points take `Iterable[KnowledgeDocument]` only where they genuinely stream. Happy to change it if you'd rather have the looser bound.

Gates: **unit 1565** (up from 1558), **integration 428**, typecheck 0, lint clean, guardrails 48.

Issue #172, opened 2026-04-14
perf: convert StatsStore/EdgeStore to persistent SQLite connections


davesnowdon:
## Problem

Every method in `StatsStore` and `EdgeStore` opens and closes a fresh `aiosqlite.connect()` connection. SQLite connection setup involves filesystem open, header read, and shared lock acquisition — this adds up when methods are called dozens of times per retrieval cycle or drain cycle.

**Examples of the pattern (repeated ~40+ times across both files):**
```python
async def get_node_stats(self, node_id: str) -> dict | None:
    await self._ensure_open()
    async with aiosqlite.connect(self.db_path) as db:
        # one query
        ...
```

The batch methods added in the MVP2 review (`get_node_stats_batch`, `increment_node_stats_batch`, `increment_coactivation_batch`) mitigate the worst hotspots on the retrieval path and full sweep, but individual methods like `update_salience`, `increment_ignored`, `get_node_stats` etc. each still open their own connection.

## Proposed Fix

Convert both stores to hold a persistent connection as a member variable:

1. Add `async open()` / `async close()` lifecycle methods (or `__aenter__`/`__aexit__`)
2. Store `self._db: aiosqlite.Connection` as a member, opened once with WAL mode
3. Replace all `async with aiosqlite.connect(self.db_path) as db:` blocks with `self._db`
4. Update server startup/shutdown to manage connection lifecycle
5. Handle connection recovery on errors
6. Update all test fixtures to use the new lifecycle

## Scope

- ~40+ method bodies in `stats.py`, ~15 in `edges.py`
- Server startup/shutdown changes
- Test fixture changes
- Changes the concurrency model — needs careful testing

## Files

- `src/lithos/lcma/stats.py`
- `src/lithos/lcma/edges.py`
- `src/lithos/server.py`
- Test fixtures in `tests/conftest.py` and individual test files

## Context

Identified during MVP2 code review on branch `ralph/mvp2-05-enrich-worker-7e9b`. The batch methods added in that review are a partial mitigation for the retrieval hot path.

---
**Lithos task:** `f0d8716d-5b9b-42a3-8ea5-7a2edc6b8363`

PR #388, opened 2026-07-15
refactor: one composition root for the component graph (PR2a, task ba8d7f25)

davesnowdon:
First of two PRs for task `ba8d7f25`. This one consolidates construction; **PR2b** takes `reconcile.py`'s disposition and the rebuild unification.

## Why

Three paths built the same component graph by hand — `LithosServer.initialize`, the `reconcile.py` per-scope functions, and half a dozen `cli` commands — and they had drifted. The cost wasn't just duplication: ADR-0006 Slice 1's **"exactly one `EdgeStore` writer"** rule was unexpressible at any interface, so `cli reindex` quietly opened a *second* writer against `edges.db` by calling `ProvenanceProjection.create(config)` with no injected store, taking its self-create branch.

## What

`lithos/pipeline.py` — `build_pipeline(config) -> Pipeline`, a frozen dataclass of the wired graph plus `aclose()`. Construction order and the one-writer rule now live once, at a named seam, **asserted by tests instead of described in comments**.

Scope is construction plus the opens it implies. Lifecycle — worker start, schema migrations, OTEL gauge registration — stays with the server, which owns process lifetime. A CLI command gets the same object graph by simply never starting the workers.

**Why Entrypoints?** It's the composition root: only `server` and `cli` import it, and Core must stay unaware of who wires it — the same reason a DI container sits at the outermost layer. Putting it in Core would mean a Core module depending on nine other Core components, which is the cycle-prone shape this model exists to catch. That placement is also edge-neutral: **`cross_component_edges` holds at 63** (module edges 131→142 collapse into `Entrypoints → Core` edges that already existed).

Adopted by `server.initialize` (**net −32 lines**; optional kwargs preserve the `if X is None` test-injection seam, so tests pre-injecting a MagicMock `search` still skip the real embedding backend) and by `cli reindex`, closing the second-writer defect.

## Also here, because it's the same surface

**Telemetry was dead on the whole CLI.** `setup_logging`/`setup_telemetry` lived in the `serve` body, so every other command ran with no-op providers — `lithos reconcile`, the operator's repair tool, emitted none of the spans and metrics its code path raises. Both now run in the `cli` group.

⚠️ **CLI surface break:** `--telemetry-console` moves from `serve` to the group. Setup now runs before any command body, so a serve-only flag couldn't be applied in time. This completes the fix — console fallback now works for `lithos --telemetry-console reconcile`, the case that motivated it.

```
lithos --data-dir ./data serve --telemetry-console   # before
lithos --data-dir ./data --telemetry-console serve   # after
```
README, `docs/cli.md`, `docs/SPECIFICATION.md` and the contract test updated.

**`extract-entities` Drift bug closed.** It called `KnowledgeManager.update` directly — skipping the Search/graph re-index, emitting no event. This is the same bug PR1a fixed on the enrich worker's path; the CLI was the other half. It now routes through `CorpusIntake.note_update(..., origin=EVENT_ORIGIN_ENRICH)`. The command used to print *"run `lithos reconcile` to refresh derived views"* — the bug admitting itself. That line is gone because there's no drift left to repair. `--dry-run` builds no pipeline, so it stays read-only and no longer risks creating `edges.db` to answer a read-only question. `ENRICH_AGENT` is re-exported via the `cognitive_memory` facade since the CLI can't cross the lcma seam.

## Test plan

- [x] `make check` — unit **1576**, typecheck 0, lint clean
- [x] `make test-integration` — **431**
- [x] `make diagrams` — guardrails **48**; `cross_component_edges` unchanged at 63
- [x] Drift regression **verified to fail against the pre-fix command**. It plants a stale entity appearing in no document's content, so the only way it leaves the index is a genuine re-index — a naive "search for the new entity" test passes either way, since entities are extracted *from* content.
- [x] New `tests/test_pipeline.py` (11): one-writer invariant, collaborator identity, injection adoption, frozen-ness, `aclose` releases the aiosqlite handle (#172), build-does-not-start-workers.

## Reviewer notes

- The `--telemetry-console` move is the one deliberate user-facing break; called out for a decision rather than buried.
- `reconcile.py` still self-creates its second `EdgeStore` (`:363`) — that's PR2b's, since it dies there anyway.
- The task described `reconcile.py` as "a shallow legacy-dict adapter". It isn't — it holds ~13 pieces of real behaviour (notably the `edges.db` short-circuit at `:351-360` that guards `ProvenanceProjection.create`'s eager file creation). That premise correction is why this is split, and why PR2b relocates rather than inlines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,14 +263,16 @@ Environment variables override `endpoint` per signal when needed:
 
 ### Local debugging without a collector
 
-Pass `--telemetry-console` to `lithos serve` to route metrics and spans to
-stdout via console exporters. This is equivalent to setting
-`telemetry.enabled=true` + `telemetry.console_fallback=true` in config, and is
-the shortest path to "is my instrumentation even firing?" when no collector is
-running.
+Pass `--telemetry-console` to `lithos` to route metrics and spans to stdout
+via console exporters. This is equivalent to setting `telemetry.enabled=true` +
+`telemetry.console_fallback=true` in config, and is the shortest path to "is my
+instrumentation even firing?" when no collector is running.
+
+It is a global option, so it works for any command — not just `serve`:
 
 ```bash
-lithos --data-dir ./data serve --telemetry-console
+lithos --data-dir ./data --telemetry-console serve
+lithos --data-dir ./data --telemetry-console reconcile
 ```
 
 ### Running the full observability stack locally
--- a/docs/SPECIFICATION.md
+++ b/docs/SPECIFICATION.md
@@ -1607,7 +1607,7 @@ lithos --data-dir ./data serve --transport http --host 0.0.0.0 --port 8765
 lithos --data-dir ./data serve --no-watch
 
 # Route OTEL metrics + spans to stdout (local debugging without a collector)
-lithos --data-dir ./data serve --telemetry-console
+lithos --data-dir ./data --telemetry-console serve
 
 # Rebuild indices (incremental by default)
 lithos --data-dir ./data reindex
--- a/docs/adr/0008-entity-extraction-ner-and-provenance.md
+++ b/docs/adr/0008-entity-extraction-ner-and-provenance.md
@@ -49,3 +49,18 @@ The v3 staging sweep surfaced two residual classes the gate still let through:
 - **Junk wiki-link targets** (`[[\phi]]`, `[[IntegerPropertyFilter(...)]]`) — wiki-link targets were kept verbatim as author intent, bypassing every other gate. v4 still keeps them lenient but drops a target carrying code punctuation (`()[]{}=<>`), a leading backslash (LaTeX), or a filename extension. Real links (`[[Knowledge Graph]]`, `[[target-doc|display]]`) are unaffected.
 
 `ENTITY_EXTRACTOR_VERSION` → 4; the `--force` re-sweep heals the corpus via the marker contract.
+
+## Amendment — `extract-entities` no longer needs a follow-up reconcile (task ba8d7f25)
+
+The Consequences above say operators run `lithos reconcile` after
+`lithos extract-entities --force` to refresh derived views. That is no longer
+true, and the requirement was never by design: the command wrote through
+`KnowledgeManager.update`, which bypassed the Search/graph re-index and emitted
+no event, so the reconcile existed to repair drift the command itself had
+manufactured (the same bypass task 681ac952 removed from the enrich worker).
+
+The command now routes through `CorpusIntake.note_update`, which re-indexes the
+derived views inline and emits `NOTE_UPDATED` — stamped `origin=enrich` so a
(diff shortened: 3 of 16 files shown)


davesnowdon:
Good catch on Finding 1 — and an embarrassing one, since enforcing that exact invariant is the whole point of the PR. Reproduced it before fixing. Both findings and both optional items are addressed in `b0d3c8e`.

## Finding 1 — fixed, and it was a family rather than a single case

Confirmed your repro. Chasing it showed the same defect from the other side: **injecting `intake` without an `edge_store`** fails identically — fresh store opened, projection built against it, injected intake left holding its own.

So rather than patch the one reported path, I did two things:

**1. Derive.** When `edge_store` is omitted, adopt an injected holder's store (projection, else intake) instead of opening a second one behind its back.

**2. Verify.** The invariant is now a **post-condition on the factory**:

```python
if projection.edge_store is not edge_store or intake.edge_store is not edge_store:
    raise ValueError("build_pipeline: projection and intake must share one EdgeStore ...")
```

Deriving fixes the cases that *can* be fixed. Contradictory injection — a projection and an intake built against genuinely different stores — cannot be, and shouldn't be papered over. This is the answer to your question: **injecting `projection` independently is supported**, and any combination that would break the one-writer rule is now rejected rather than silently wired. That's what makes the factory *enforce* the invariant instead of merely expressing it, which is the point of the PR.

Three new tests, all **verified to fail against the pre-fix factory**:
- `test_injected_projection_supplies_the_edge_store` — your exact case
- `test_injected_intake_supplies_the_edge_store` — the same defect from the other side
- `test_contradictory_injection_is_rejected`

Also adds **`CorpusIntake.edge_store`** so the composition root can check the invariant without reaching through `_edge_store`. It mirrors the accessor added to `ProvenanceProjection` in PR1b for exactly this reason, and it let the tests drop their own private access.

## Finding 2 — fixed

Right, and the ADR was wrong in a way worth recording rather than just deleting: that reconcile requirement was **never by design**. It existed to repair drift the command manufactured by bypassing intake — so the ADR had faithfully documented a bug as a procedure. Amended in ADR-0008's existing amendment style, noting extraction is still a corpus mutation (still not a Reconcile per CONTEXT.md); what changed is that it no longer leaves drift behind for one.

## Nice-to-have — done

`Pipeline.aclose()` now documents ownership: it closes the store regardless of who built it, so a caller that injected one owns that store's lifecycle. Mirrors the wording `ProvenanceProjection.close` already uses. I kept it documented rather than tracked (no `_owns_edge_store` flag) because ownership is unambiguous in practice — production callers pass `config` alone, and injection is a test affordance. Happy to make it explicit state if you'd rather.

Gates: **unit 1579** (up from 1576), **integration 431**, typecheck 0, lint clean, guardrails 48. `cross_component_edges` still 63.

PR #402, opened 2026-07-23
feat(lcma): recalibrate salience — decay-to-floor + usage signal (e7d8ef60)

davesnowdon:
## What & why

Salience (LCMA per-node retrieval-utility signal) had **collapsed in prod**: avg **0.076**, **86% of ~1,981 nodes ≤ 0.30**. Root cause, confirmed in code:

- Every node starts at 0.5.
- A **floor-less daily sweep** decays untouched nodes linearly toward **zero** (`enrich.py` `_apply_decay` / `full_sweep`).
- The only forces that *raise* salience fire on actively cited/consolidated nodes — never the cold bulk corpus, and near-empty in prod.

So the rerank salience term (`+ salience*0.1`, raw/unnormalized) became a near-uniform ~0 penalty — degrading **all** `lithos_retrieve` ranking and blocking MVP-3 WS4 (concept damping via a salience ceiling) and WS6 (salience-weighted rerank), both of which assume a meaningful spread. Fixes tracker task `e7d8ef60` (the MVP-3 WS0 prerequisite gate).

## Approach

1. **Decay to a non-zero floor** — time decay stops at `lcma.salience_floor` (default 0.3); explicit negative feedback (misleading / chronic-ignored) may still go below. Math in a new pure module `lithos/lcma/salience.py`.
2. **Usage signal** — a new non-decaying `usage_score` rerank term (log-scaled retrieval frequency + exponential recency), computed from counters the rerank pre-fetch **already** reads (`get_node_stats_batch` does `SELECT *`) → **zero extra queries**. Separates learned *quality* (stored salience) from live *popularity* (usage) and cannot collapse. Emitted as a new result field.
3. **Config exposure** — the previously-hardcoded decay/reinforcement constants and the composite weights become `LcmaConfig` fields (also gives WS6 its tuning surface).
4. **One-time backfill** — `lithos recalibrate-salience` (idempotent, staging-first), routed through the `cognitive_memory` facade to respect the lcma import seam. Lifts collapsed rows to the floor; skips explicit-negative-feedback nodes.
5. **Observability** — the daily sweep snapshots the salience distribution (mean, fraction **strictly below the configured floor**, node count) to OTEL gauges so a re-collapse is alertable. The below-floor fraction is the collapse marker: it clears to ~0 once the backfill parks rows *at* the floor (nodes resting at the floor are healthy), which a fixed `≤ 0.30` threshold could not distinguish.
6. **Calibration harness** — `scripts/calibrate_salience.py`, an offline eval that replays a stats.db copy through the same pure helpers, reports distribution health + separation (citation/activity AUC), and ranks parameter sets. Sets the shipped defaults.

Shipped defaults are **provisional** pending calibration against a prod/staging stats.db copy and the operator backfill run.

## Decisions (agreed before implementation)

- Scope = floor **+** usage signal (floor-only leaves every cold node piled at the floor).
- Full calibration harness (feeds WS6 later).
- Backfill = operator-run CLI, staging-first (keeps stats.db migrations DDL-only; data mutation stays explicit).

## Test plan

- `make check` (lint + typecheck + unit) and `make test-integration` — **green**.
- New coverage: pure helpers + eligibility predicate (`test_salience.py`), decay-floor + config wiring + gauge emission (`test_enrich_worker.py`), usage-in-rerank + config wiring (`test_retrieve.py`), backfill + floor-aware distribution + percentile + floor validation (`test_stats.py`), cited-boost wiring (`test_cognitive_memory.py`), CLI (`test_cli.py`), harness incl. end-to-end receipt parsing + time-split (`test_calibrate_salience.py`). lcma import-seam guard still green.
- Operational, **post-merge**: run the harness against a stats.db copy to finalize the provisional defaults, then `recalibrate-salience` on **staging → prod**, watching the new distribution gauge.

## Not in scope

- Dropping the vestigial `decay_rate` column (a NOT-NULL rebuild on live prod stats.db + a `lithos_node_stats` surface change) — left as low-priority cleanup.

### DIFF
--- a/docs/SPECIFICATION.md
+++ b/docs/SPECIFICATION.md
@@ -1134,7 +1134,8 @@ PTS-style retrieval orchestrating seven scouts in parallel and reranking via a f
       "derived_from_ids": [],
       "reasons": ["lexical match score 0.91"],
       "scouts": ["scout_lexical"],
-      "salience": 0.42
+      "salience": 0.42,
+      "usage_score": 0.18
     }
   ],
   "temperature": 0.5,
@@ -1145,7 +1146,7 @@ PTS-style retrieval orchestrating seven scouts in parallel and reranking via a f
 }
 ```
 
-The `id`/`title`/`snippet`/`score`/`path`/`source_url`/`updated_at`/`is_stale`/`derived_from_ids` keys mirror the `lithos_search` shape so clients that read only those fields work unchanged. `reasons`/`scouts`/`salience` are LCMA-only additive fields. `degraded`/`failed_scouts` are always present: `failed_scouts` lists the canonical names of any scouts whose backend raised (so one bad backend degrades rather than kills the retrieve), and `degraded` is `true` when that list is non-empty — letting a caller distinguish partial results from a genuinely empty corpus.
+The `id`/`title`/`snippet`/`score`/`path`/`source_url`/`updated_at`/`is_stale`/`derived_from_ids` keys mirror the `lithos_search` shape so clients that read only those fields work unchanged. `reasons`/`scouts`/`salience`/`usage_score` are LCMA-only additive fields — `salience` is the stored learned utility (`stats.db`), while `usage_score` is the live, non-decaying popularity signal (retrieval frequency + recency) that feeds reranking alongside it. `degraded`/`failed_scouts` are always present: `failed_scouts` lists the canonical names of any scouts whose backend raised (so one bad backend degrades rather than kills the retrieve), and `degraded` is `true` when that list is non-empty — letting a caller distinguish partial results from a genuinely empty corpus.
 
 **Receipt audit trail:** every call writes a row to `stats.db.receipts` with columns including `id`, `ts`, `query`, `namespace_filter`, `scouts_fired` (canonical names of every scout that ran cleanly — empty results still count as fired), `candidates_considered`, `final_nodes` (JSON array of `{id, reasons, scouts}` objects), `surface_conflicts`, `temperature`, and `terrace_reached`.
 
--- a/docs/architecture.toml
+++ b/docs/architecture.toml
@@ -115,7 +115,7 @@ module_cycles          = 1    # server<->tools; the knowledge<->graph<->search<-
 modules_over_800_lines = 11   # incl. tools.tasks (985) split from server in #374; knowledge.py 2444→1773 (codec out) but still over; direction: down
 max_module_lines       = 2800 # stop-loss; largest today is coordination.py at 2675
 cross_module_private_refs = 42  # measured 2026-07; server._emit/_config x12 etc. (normalize_datetime made public in task 4a3836a9); goal: 0
-tests_private_imports     = 88  # measured 2026-07; test_telemetry alone is 45; direction: down
+tests_private_imports     = 90  # measured 2026-07; test_telemetry alone is 45; direction: down. 88→90: two white-box config-wiring tests import _rerank_fast / _usage_from_stats to prove the newly-exposed LcmaConfig rerank/usage fields actually change behaviour (task e7d8ef60, review finding 5) — the private helpers are the precise units under test.
 
 # MCP tool surface: modules scanned for @…tool()-decorated handlers (counted in
 # the metrics snapshot; catalogued in docs/generated/tool_catalog.md). Omit the
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -113,6 +113,27 @@ lithos reindex --clear
 
 Use this after manually editing or adding Markdown files outside of the MCP interface.
 
+### `recalibrate-salience` — One-time salience backfill
+
+```bash
+# Preview: show the distribution and how many rows would be lifted
+lithos recalibrate-salience --dry-run
+
+# Lift decay-collapsed rows up to the floor (config lcma.salience_floor)
+lithos recalibrate-salience
+
+# Override the floor explicitly
+lithos recalibrate-salience --floor 0.3
+```
+
+Corrects the historical salience collapse (task `e7d8ef60`) where a floor-less daily
+decay drove most nodes to ~0. Lifts every node below the floor up to it, **except** nodes
+carrying explicit negative feedback (misleading / chronically ignored), which stay below
+deliberately. Idempotent — safe to re-run. Run it **once on staging first, then prod**,
+with the server stopped is not required (it only touches `stats.db`), but the operator
+should confirm the reported before/after distribution. Going forward the daily enrich
+sweep holds the floor, so this should not need re-running.
+
 ### `validate` — Check knowledge base integrity
 
 ```bash
--- a/docs/generated/components/CognitiveMemory.md
+++ b/docs/generated/components/CognitiveMemory.md
@@ -11,13 +11,14 @@ Agent-facing facade of LCMA: retrieval, learning, and asserted state.
 
 | Module | Size | Classes | Functions |
 |---|---|---:|---:|
-| `lithos.cognitive_memory` | L | 2 | 0 |
+| `lithos.cognitive_memory` | L | 2 | 1 |
 
 ## Public API
 
 ### `lithos.cognitive_memory`
 - class `NodeStats` — Per-node retrieval / salience stats returned by :meth:`CognitiveMemory.node_stats`.
 - class `CognitiveMemory` — Module facade over LCMA's agent-facing surface (ADR-0005).
+- def `recalibrate_salience` — Run the one-time salience floor backfill as a standalone maintenance op.
 
 ## Dependencies
 
--- a/docs/generated/components/Entrypoints.md
+++ b/docs/generated/components/Entrypoints.md
@@ -11,7 +11,7 @@ CLI, MCP server, tool handlers, filesystem watcher — the outward-facing surfac
 
 | Module | Size | Classes | Functions |
 |---|---|---:|---:|
-| `lithos.cli` | L | 0 | 15 |
+| `lithos.cli` | L | 0 | 16 |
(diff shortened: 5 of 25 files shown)


davesnowdon:
Thanks both — this was a genuinely useful review. Follow-up commit addresses every finding; summary below.

## Agent review

**1. `--floor` could persist salience outside [0,1] — fixed.** CLI `--floor` is now `click.FloatRange(0.0, 1.0)`; `StatsStore.recalibrate_salience_floor` raises `ValueError` on any non-finite / out-of-range floor (defense in depth). Tests cover negative, >1, NaN, and inf at both the CLI and store layers.

**2. Harness didn't reproduce the real backfill — fixed.** Eligibility is now a single pure predicate `lithos.lcma.salience.recalibration_eligible(...)`, used by both the harness projection and (mirrored as SQL) the store. `NodeRow` loads `misleading_count` / `ignored_count` / `cited_count`, and `floor_report` excludes protected nodes exactly as the operator run does. New test seeds collapsed / misleading / chronic-ignore / light-ignore / high rows and asserts the harness's projected lift set == the store's actual lift set.

**3. Retrieval-activity AUC was circular — replaced.** The circular fallback is gone. The harness now computes a **non-circular future-retrieval AUC** from the `receipts` log: it time-splits receipts (default 70/30), scores each node from its **past** retrievals only, and labels it by whether it's retrieved again in the held-out **future** window — so the label is not an input to the score. Ranking prefers citation-AUC (independent), then future-retrieval-AUC; when neither label has ≥10/≥10 support (the sparse regime), the harness prints distribution diagnostics **and explicitly does not claim a best config**. New test proves the metric is informative: with frequency held constant, a recency-aware config gets future-AUC 1.0 while a frequency-only config gets 0.5, and the informative one ranks first.

**4. Collapse gauge couldn't tell recovery from resting-at-floor — fixed.** `salience_distribution(floor)` now reports `fraction_below_floor` (**strictly** below the configured floor) instead of a fixed `<= 0.30`. It's high while collapsed and clears to ~0 once the backfill parks rows *at* the floor. Gauge renamed `lithos.lcma.salience.fraction_below_floor`; the sweep passes `lcma.salience_floor`. New test asserts the signal is 1.0 before and 0.0 after a backfill.

**5. Config-wiring not proven — added.** Decisive tests that a non-default value changes behaviour: decay `floor`/`per_day`/`cap` (a custom-config worker stops decay at 0.45, not 0.30); `usage_recency_halflife_days` (longer half-life ⇒ higher usage on a 14-day-old use); `rerank_usage_weight` (at 0.1 the high-usage node wins; at 0.0 the tie is stable); `salience_cited_boost` (custom 0.1 ⇒ salience 0.6, not 0.52). (This nudged `tests_private_imports` 88→90 — two white-box tests import the precise private helpers under test; bumped with a justification in `architecture.toml` per the guardrail's reviewable path.)

**6. Design doc stale — updated.** `lcma-design.md` result shape now lists `usage_score`, the `LcmaConfig` draft schema carries the new floor/decay/reinforcement/usage/weight fields (with a pointer to `config.py` as authoritative), the response-compat section includes `usage_score`, and the `confidence` vs `salience` note now distinguishes `usage_score` as live popularity vs learned quality.

## Copilot

- **Percentile off-by-one — fixed.** Nearest-rank now `ceil(q·n)` clamped to `[1,n]`; `p50` on `n=2` picks the lower value. Test added.
- **`_days_since` could raise on a bad timestamp — fixed.** Wrapped in `try/except ValueError` → returns `None` (recency term drops out) rather than failing the retrieval.
- **`create_gauge` availability.** Verified present in the pinned OTEL SDK (synchronous `Gauge` is stable API since 1.28), and `_NoOpMeter` now implements it for the disabled path. I also made the sweep's emission fully defensive so any telemetry hiccup logs-and-continues rather than aborting the sweep's decay/WM/provenance work.
- **Mixed import style — fixed** via `sys.modules[EnrichWorker.__module__]`.

## Answers to your questions

1. **AUC for defaults or sanity check?** Selection now uses an *independent* label (citations, or held-out future retrieval); with no supported label the harness is descriptive-only and picks nothing. The independent outcome replacing retrieval-count is future retrieval.
2. **Fixed threshold or configured floor?** Configured floor. `fraction_below_floor` (strict `<`) distinguishes "collapsed below floor" from "healthy resting at floor."
3. **Does `misleading_count > 0` permanently protect all sub-floor salience?** Yes, deliberately. The backfill is a one-time decay-repair and can't attribute how much of a low value came from the single feedback hit vs decay; since explicit feedback is near-empty in prod (the `fc4b0669` finding) the protected set is tiny, and erring toward *preserving* a deliberate penalty beats erasing it. If feedback becomes dense we can narrow protection to the quarantine threshold — noted as a follow-up.
4. **Popularity loop vs concept damping.** Real — flagged as a risk in the plan. WS4's salience ceiling won't cap a concept node's `usage_score`; that term will need its own cap/tempering in the WS4 design. Present mitigations: log-scaled frequency + recency decay (a result must keep being used) + MMR diversity, and the harness's saturation check.
5. **Exposure as utility, or tempered?** `usage_score` is deliberately raw exposure, kept *separate* from learned quality (`salience`) and combined additively with an independent weight — so `rerank_usage_weight` already lets an operator temper popularity, while citations/feedback flow through `salience`. Recency decay is the main entrenchment guard; monitoring result diversity (as you suggest) is a good follow-up, and WS6 metamemory can eventually learn the balance.

## Deliberately not in scope

Dropping the vestigial `decay_rate` column (a NOT-NULL rebuild on live prod `stats.db` + a `lithos_node_stats` surface change) stays a low-priority cleanup.

`make check` (unit + lint + typecheck) and the integration suite are green on the follow-up.

davesnowdon:
Thanks — good catch on the receipt shape. Both findings fixed.

**1. Receipt parsing (merge-blocker) — fixed.** `final_nodes` is a list of `{"id", "reasons", "scouts"}` objects, and my loader was stringifying each dict, fragmenting a node by its explanation payload. The harness now reuses the store's own parser, `lithos.lcma.stats._extract_final_node_ids`, so it reads receipts *exactly* as production does and can't drift from it. Added an end-to-end test that inserts production-shaped receipts via `StatsStore.insert_receipt()`, loads them through `load_receipts`, and asserts two objects for the same node (different `reasons`) collapse to the single id `["node-1"]` — not two pseudo-identifiers.

**2. Split off-by-one — fixed.** `build_time_split` now slices by *position* (`ordered[:past_count]` / `ordered[past_count:]`) over `(ts, rowid)` order, validates `0 < split < 1`, and returns `None` for degenerate inputs (fewer than two receipts). Test asserts a 0.7 split of 10 receipts puts exactly 7 in the past (n0–n6) and 3 in the future, and that 0.0 / 1.0 / single-receipt inputs fall back to descriptive-only.

**`Z`-suffix robustness (Copilot) — fixed.** `_days_since` (retrieve) and the harness `_parse_ts` now normalise a trailing `Z` to `+00:00` before `fromisoformat`, matching `server.py` / `coordination.py`, so an alternate/older stored timestamp parses instead of silently dropping the recency term. Covered by tests (`Z`-suffixed → higher usage than long-ago; garbage → dropped, not raised).

**Test-plan text — updated.** The PR description now reflects the green integration run.

## Answers

1. **Run against a real stats.db with production `final_nodes` yet?** Not yet — that's the post-merge operational step (run the harness on a prod/staging `stats.db` copy to finalize the provisional defaults). The new end-to-end test closes the coverage gap in CI by exercising the real `insert_receipt` → `load_receipts` path with the production object shape, so we're no longer relying on a hand-built `TimeSplit`.
2. **Support legacy plain-string receipt IDs?** No — production has only ever written the object shape (`retrieve.py` builds `{"id", ...}` dicts), and by reusing the store's `_extract_final_node_ids` the harness is guaranteed consistent with the authoritative reader. No string-shaped receipts exist to support.
3. **Position-based split with ts/rowid ordering?** Yes — done exactly that: order by `(ts, rowid)`, split by position, so receipts sharing a second-resolution timestamp no longer all land in the past.

`make check` (unit + lint + typecheck) and the integration suite are green on the follow-up.

davesnowdon:
Thanks for the approve. Addressed all three minor findings.

**1. Observability wording — fixed.** PR description now says "fraction **strictly below the configured floor**" (with the note that a fixed `≤ 0.30` threshold couldn't distinguish recovery from resting-at-floor). Confirmed: `< configured floor` is the intended long-term contract — nodes parked at the floor by the backfill are healthy, so the marker clears on recovery.

**2. Equal-timestamp ordering — regression-tested.** New test inserts two receipts for *distinct* nodes, forces an identical `ts`, and asserts the past/future partition follows rowid (insertion) order — i.e. the `(ts, rowid)` contract holds and same-second receipts split deterministically. It would fail if the query or sort dropped the rowid tie-break.

**3. Private parser — promoted.** `extract_final_node_ids` is now a public package helper (single implementation, shared by the store's receipt readers and the harness), so an internal refactor can't silently break the operator harness.

## Answers

- **Is `< configured floor` the intended alert contract?** Yes — updated the description accordingly.
- **Are all receipt rows object-shaped?** Yes. Every version of the pipeline writes `final_nodes` as `{"id", "reasons", "scouts"}` objects (`retrieve.py`), and the authoritative reader (`extract_final_node_ids`) only handles that shape — the harness now reuses it, so there's no code path that ever wrote or reads string arrays. A `SELECT final_nodes FROM receipts LIMIT 5` against the staging copy will confirm empirically when the harness is run there.
- **Operational owner for the harness + staging-first backfill?** That's the operator's call (you) per the plan — I'll run the harness on a `stats.db` copy and the staging→prod backfill under your direction, but won't touch staging/prod or self-merge without your go-ahead. Happy to also record the calibrated parameters + dataset characteristics as a KB note after the run (your "nice to have").

`make check` and the integration suite are green on the follow-up.

Issue #306, opened 2026-06-03
Add metadata-equality filtering to lithos_list and lithos_task_list (#305 Part B)


davesnowdon:
> *This was generated by AI during triage.*

## Summary

Follow-up to #305 (Part B). Part A added writable + readable key-value `metadata` on the document write surface (persisted via `KnowledgeMetadata.extra`, returned by `lithos_read`/`lithos_list`). This issue covers the deferred **metadata-equality filtering** on the list APIs.

Today neither `lithos_list` nor `lithos_task_list` can filter by metadata, even though both surfaces now carry writable free-form metadata. A consumer that wants "every doc where `github_watch_enabled == true`" must iterate and filter client-side.

## Proposed change

Add a symmetric `metadata_match: dict` equality filter to **both** list tools, designed identically so the two surfaces share one consistent filter shape:

- `lithos_list(..., metadata_match={"k": "v"})` → returns only docs whose `extra` equals the given value on each supplied key. Slots into the Python-side post-filter in `KnowledgeManager.list_all()` against the cached metadata (`extra`), no disk I/O.
- `lithos_task_list(..., metadata_match={"k": "v"})` → symmetric, applied as a post-SQL Python filter against the decoded task `metadata` in `CoordinationService.list_tasks()`.

Equality-only (no ranges/operators) for v1.

## Acceptance

- `lithos_list` returns only docs matching every key/value in `metadata_match`.
- `lithos_task_list` supports the same filter with identical semantics.
- Tests cover match, partial-key match, and no-match for both surfaces.
- `make check` + `make test-integration` green.

## Context

- Part A PR introduced the writable/readable metadata surface and the shared `lithos._merge.merge_metadata` helper.
- The original #305 body marks doing both parts "the stronger move" for API consistency; this issue is the second half.

PR #411, opened 2026-08-11
fix(codec): bare YAML dates in frontmatter break indexing (#407)

### LINKED ISSUE #407: v0.4.0: bare date in frontmatter breaks indexing — "Object of type date is not JSON serializable" (regression)
## Summary

On **v0.4.0**, any knowledge note whose frontmatter contains a **bare (unquoted) date** — e.g. `created: 2026-07-30` — is skipped during indexing with `Object of type date is not JSON serializable`, and the Tantivy backend fails to build. This worked on v0.2.x, so it's a regression. In our vault it silently dropped **746 of 2975 notes (~25%)** and left **fulltext search returning zero results** (`stats` showed `Search chunks: 0`).

## Environment

- lithos **v0.4.0** (tag, commit `11a3e30`)
- Python 3.12, stock `docker/Dockerfile` image

## Reproduction

1. Create `note.md` with an **unquoted** date in a non-reserved key:
   ```markdown
   ---
   id: repro-1
   created: 2026-07-30
   ---
   hello
   ```
   (PyYAML parses bare `2026-07-30` as `datetime.date`.)
2. `lithos reindex`

Observed:
- `stats` logs `Skipping invalid file …/note.md: Object of type date is not JSON serializable`
- reindex logs `Failed to repair tantivy backend: sequence item N: expected str instance, datetime.date found`
- the note is **not indexed**; fulltext returns nothing for its content.

Quoting the value (`created: '2026-07-30'`) loads it as `str` and it indexes fine — that's the workaround we used across the vault.

## Root cause

`extract_extra()` (`src/lithos/knowledge.py:237`) sweeps every non-reserved frontmatter key into free-form `extra` (#305). Reserved date fields (`created_at`, `updated_at`) are `.isoformat()`-ed, but `extra` values are passed through as-is. When such a value becomes a canonical bucket key for `metadata_match` (#306):

```python
# src/lithos/knowledge.py:246-252
def canonical_metadata_value(value: object) -> str:
    return json.dumps(value, sort_keys=True, separators=(",", ":"))
```

`json.dumps` has no `default=`, so a `datetime.date` / `datetime.datetime` value raises `TypeError: Object of type date is not JSON serializable`. The Tantivy indexing path hits the same class of failure (`expected str instance, datetime.date found`).

Since YAML natively parses `YYYY-MM-DD` as `datetime.date`, **any** note with a bare date under a non-reserved key (`created:`, `date:`, `updated:`, `published:`, …) triggers it — extremely common in agent-authored vaults.

## Suggested fix

Normalize date/datetime in the `extra` path so the whole indexing chain (canonical key + Tantivy + Chroma) sees strings. Minimal:

```python
return json.dumps(value, sort_keys=True, separators=(",", ":"), default=str)
```

or coerce `datetime.date` / `datetime.datetime` → `.isoformat()` inside `extract_extra()` when building the indexable metadata.

## Impact / severity

The failure is **silent**: notes are skipped with a per-file log line, not a hard error, so fulltext quietly degrades toward empty while `/health` still reports the KB present. A regression test with a bare-date frontmatter fixture would catch it.

davesnowdon:
Fixes #407.

## The regression

PyYAML resolves an unquoted `created: 2026-07-30` to `datetime.date`. Any note carrying one under a non-reserved key (or in `tags:`) breaks the indexing chain:

- `canonical_metadata_value` (`json.dumps` with no `default=`) raises `TypeError: Object of type date is not JSON serializable`. On v0.4.0 that silently skipped the note (~25% of the reporter's vault vanished from search); since the CorpusIndex extraction it is worse — the exception escapes `rebuild()` and **crashes `KnowledgeManager.__init__`, so the server won't boot** (the per-file try only guards file parsing, not index building).
- A bare date in `tags:` reaches `" ".join(doc.tags)` in the Tantivy backend: `sequence item N: expected str instance, datetime.date found` — one bad doc poisons the whole `rebuild_from_docs` commit, leaving fulltext empty.

## The fix

Normalize at the ingestion boundary, once:

- New `normalize_yaml_dates()` in `frontmatter_codec.py` recursively converts YAML-native `date`/`datetime` values to ISO-8601 strings (lists/dicts included). Called at the top of both YAML ingestion points — `KnowledgeMetadata.from_dict` and `CachedMeta.from_frontmatter` — so a bare date behaves exactly like its quoted form everywhere downstream: JSON bucket keys, `metadata_match` filtering, Tantivy/Chroma feeds, MCP response serialisation, and round-trip writes (first re-encode quotes the value; thereafter a stable fixed point).
- Backstop: `canonical_metadata_value` now passes `default=str`, making it total — it runs during the startup index rebuild, where raising on any exotic hand-edited value (e.g. `bytes`) would stop the server from booting. A lossy-but-deterministic bucket key is strictly better than that.

`frontmatter_codec.py` stays under the 800-line budget (799).

## Tests

Regression coverage at all three layers (12 new tests):

- **Codec** (`test_codec_conformance.py`): bare date / bare datetime / nested lists+dicts decode to strings; bare date in `tags` decodes to strings (the Tantivy join vector); bare ≡ quoted equivalence; round-trip idempotence; `canonical_metadata_value` totality.
- **Corpus index** (`test_corpus_index.py`): `from_frontmatter` stringifies; a dated note indexes and matches its string query form; `rebuild` survives bare-date frontmatter.
- **Knowledge** (`test_knowledge.py`): the end-to-end repro — an on-disk note with `created: 2026-07-30` and a dated tag boots the manager, is present in the index, matches `metadata_match`, and produces a joinable `IndexableDocument`.

## Verification

- `make check`: 1782 unit tests passed, lint + typecheck green
- `make test-integration`: 470 passed
- `make diagrams` regenerated (LOC metrics + Codec component page committed)


davesnowdon:
Both findings confirmed and fixed in 7968501 — thanks, the nested-key gap was real and I reproduced both crashes end-to-end through `KnowledgeManager` before fixing.

**Finding 1 (nested date keys / mixed key types)** — fixed at both layers:

- `normalize_yaml_dates` now normalizes mapping **keys** as well as values (`{_norm(k): _norm(v)}`), so `custom: {2026-07-30: value}` ingests as `{"2026-07-30": "value"}` and never reaches a failing canonicalization path.
- `canonical_metadata_value` is now total by construction: `default=str` covers unserialisable *values*, and a `TypeError → repr(value)` fallback covers structures JSON cannot encode or sort at all (date keys arriving via any path, mixed `str`/`int` keys). `repr` of a parsed document is deterministic (dict insertion order = YAML file order).

**Answers to your questions:**

1. *Is `canonical_metadata_value()` intended to be total for every structure PyYAML can produce?* — Yes, and now it genuinely is: the JSON path handles everything common, and the `repr` fallback catches the remainder, so the startup rebuild can never raise from here. The key observation that makes the fallback safe: `validate_metadata_match` restricts query values to JSON scalars, so **dict-shaped stored values are unreachable by matching** — their bucket key only has to be stable, not JSON-faithful. This is now stated in the docstring.

2. *Collision behavior for differently-typed keys with the same string representation?* — Ingestion coercion is deliberately **type-aware, not naive**: only `date`/`datetime` keys (the YAML-vs-JSON type mismatch this bug is about) are stringified. A date key colliding with its quoted twin resolves last-wins in file order (deterministic; tested). `1` vs `"1"` are *not* coerced — they remain distinct Python keys, and the `repr` fallback preserves that distinction, so no collision policy is needed for them at all.

**Tests added** (per your list): nested mapping with a bare date key; quoted-twin collision policy; canonical totality + determinism for mixed `str`/`int` keys and date keys; corpus-index `rebuild` and full `KnowledgeManager` startup with both structures.

**One budget note for the maintainer:** the fix landed in a module I had already trimmed to 799/800, so `modules_over_800_lines` is bumped 11→12 with a reviewed comment in `architecture.toml` naming the follow-up split candidate (the #306 metadata-value cluster) and why it's blocked today (`_KNOWN_METADATA_KEYS` lives beside `KnowledgeMetadata`; moving it creates an import cycle).

Gate: `make check` green (1784 unit), `make test-integration` green (474), diagrams regenerated.

PR #412, opened 2026-08-26
feat(api): accept unambiguous short ID prefixes for task/note IDs; echo resolved id+title

davesnowdon:
Implements Lithos tracker task `83257ced-2469-4af1-8c9d-290acf5e9854`: every human-facing artefact refers to tasks/notes by 8-char prefix, but the API took only full UUIDs — and a rejected short ID pushes agents into *reconstructing* UUIDs from context, where a fabricated UUID that happens to exist silently writes to the wrong task. This closes that hallucination surface at the parameter boundary.

## What

1. **Unambiguous short prefixes (min 6 chars, git-style) accepted wherever a task/note id is taken** — the task lifecycle set (get/update/claim/renew/release/complete/cancel/reopen/status/children/spawn), task edges (upsert both endpoints, list), `depends_on`/`parent_task_id`/`source_task_id` on create/spawn, findings (post/list), and the note side (read/write/note_update/delete/related/node_stats). Task prefixes resolve against tasks only, note prefixes against notes only — no unified namespace.
2. **Ambiguity fails loudly**: new `ambiguous_id_prefix` code carrying up to 5 `{id, title}` candidates — never a silent pick.
3. **Every mutating response echoes the resolved full id + title** (`task_id`/`title` on task mutations, endpoint ids+titles on edge upsert, `title` on write/note_update, `id`/`title`/`path` on delete) — writes are self-verifying and transcripts retain full ids.

## How

- **Golden rule: full-length (36-char) task ids pass through untouched** → zero behavior change for existing callers (`claim_failed`, `claim_not_found`, silent-empty lists, orphan findings all preserved). Notes are exact-match-first at ANY length (hand-authored notes may carry arbitrary string ids), then the same loud prefix path.
- New pure Foundation module `lithos.id_resolution` (`MIN_PREFIX_LEN`, `prefix_upper_bound`, `PrefixIndex`) shared by both resolvers; `AmbiguousIdPrefixError` lives in `lithos.errors` and is mapped unconditionally in the `tool_span` seam (one change covers all 37 tools).
- **Sub-linear on both sides** (project rule): tasks use a half-open PK range scan — `LIKE 'p%'` full-scans under the id column's BINARY collation — **pinned by an EXPLAIN QUERY PLAN test** asserting SEARCH-not-SCAN; notes use a sorted id mirror (bisect) maintained at every `_id_to_path` mutation incl. the watcher `set_path` path.
- Resolution is tool-boundary only: internal callers (LCMA, intake, watch, CLI) keep exact-match semantics.

## Deliberate wire-contract changes (pinned in tests + CHANGELOG)

- <6-char ids matching nothing exactly → `invalid_input` (were `task_not_found`/`doc_not_found`/`claim_failed`/silent).
- 6–35-char task ids matching nothing → `task_not_found` on every task tool, including claim/renew/release and the formerly silent-empty status/edge_list/children/finding_list.
- **Bug fix**: `lithos_write` with an unknown `id` returned an uncaught `FileNotFoundError` → protocol-level ToolError; it is now a proper `note_not_found` envelope.
- New echo keys are additive.

## Architecture / guardrails

- `docs/architecture.toml`: new `IdResolution` Foundation component; `cross_component_edges` 72→74 (reviewed comment: Coordination→IdResolution, Knowledge→IdResolution). Both import-linter Foundation contracts gained `lithos.id_resolution`. No `[domain]` change (module is dataclass-free by design).
- Budgets hold without further edits: coordination.py 2730/2800 stop-loss, corpus_index.py 782/800, read_search.py 745/800.
- `docs/generated/` regenerated (new IdResolution component page); SPECIFICATION gains §10.3 "Short ID Prefixes" + updated Returns lines; CHANGELOG entry under Unreleased.

## Tests

- `tests/test_id_resolution.py` (new): `prefix_upper_bound` incl. dash-terminal + carry cases; `PrefixIndex` semantics.
- `tests/test_coordination.py::TestResolveTaskId`: passthrough, prefix hit/miss/too-short/ambiguous (candidate cap), and the **EXPLAIN QUERY PLAN index-search pin**.
- `tests/test_corpus_index.py` / `test_knowledge.py`: prefix mirror across add/remove/rebuild/watcher-set_path; `resolve_id` exact-first (incl. a 3-char hand-authored id), ambiguity with ordered candidates.
- `tests/test_envelope_conformance.py`: new VALIDATION_CASES (too-short → `invalid_input`; claim prefix-miss → `task_not_found`) + golden ambiguity envelope with candidates; the exact not-found pin updated to a full-length ghost id (5-char "ghost" now deliberately reports `invalid_input`).
- `tests/test_id_prefix_tools.py` (new, 13 e2e): prefix acceptance + echo across the surface; `lithos_write` unknown-id → `note_not_found` regression pin.
- `tests/test_tool_seam.py`: ambiguity mapping under both `map_coordination_error` values.
- Stale exact-payload pins updated in 6 integration files + test_task_complete_feedback/test_task_graph.

## Test plan

- [x] `make check` green (typecheck + 1818 unit tests + lint)
- [x] `make test-integration` green (491 tests)
- [x] `make diagrams` drift committed (metrics, component pages, tool catalog)


davesnowdon:
All five findings addressed in 8f98d3d — thank you, the probes were exactly right: these boundaries could persist or filter by a display prefix, which is the silent-wrong-state class this PR exists to close.

One deliberate deviation from a literal reading of the review: these are all **reference/context fields whose contracts allow values that are not (yet) records** — `derived_from_ids` forward references, free-form asserted-edge node ids (`concept:foo`), findings posted ahead of their note, and `retrieve.task_id`/`write.source_task` doubling as free-form correlation keys (the existing LCMA suites pin values like `task-wm-test`). Strict resolution would break all of those. So every fixed surface uses **lenient resolution**: an exact or unique-prefix hit resolves to the full id, ambiguity fails with `ambiguous_id_prefix`, and any other value passes through unchanged. §10.3 now documents the strict-vs-lenient split explicitly, so the "universal contract" no longer overclaims.

Per finding:

1. **retrieve.task_id** — resolved via new `CoordinationService.resolve_task_id_lenient` before `memory.retrieve`; your probe is now a regression test (`test_retrieve_task_prefix_matches_full_id_scoping`: a task-scoped note visible with the full id is equally visible with its 8-char prefix), plus a passthrough pin for correlation-key usage.
2. **write.source_task** — resolved with the same lenient task resolver before the request is built; pinned by `test_write_source_task_persists_full_id` (frontmatter `source` carries the full id) and a free-form passthrough pin.
3. **finding knowledge_id** — resolved via new `KnowledgeManager.resolve_id_lenient`; prefix links now store the full note id (pinned), and an unknown full UUID still stores verbatim (findings-ahead-of-note pinned).
4. **asserted edges** — `edge_upsert` endpoints, `edge_list` filters, and `conflict_resolve.winner_id` all resolve leniently; pinned that a prefix-created edge is found by full-id queries (and by prefix filters), and that free-form node ids keep working. Ambiguity coverage lives at the resolver unit level plus the seam's unconditional mapping test.
5. **derived_from_ids** — each entry resolves leniently *before* UUID validation, so a prefix of an existing note resolves while forward references stay full UUIDs (pinned, including the mixed case).

Gates: `make check` (1823 unit) and `make test-integration` (500) both green at 8f98d3d; metrics snapshot regenerated in the same commit.

davesnowdon:
Re the P2: correct observation — the spec and the resolver disagreed. Fixed in 34e6755 by taking the **narrow-the-contract** option.

Rationale for narrowing rather than dual-namespace resolution: the only consumer of the stored `source` value anywhere in the codebase is the task-scope access gate (`_passes_access_scope`, scouts.py — `source == task_id`). No code path reads `source` as a note reference, and note provenance already has a dedicated, resolved field (`derived_from_ids`). Dual-namespace resolution would require a task-vs-note prefix collision policy for a semantics nothing implements. The spec's "or provenance note" clause was legacy wording.

Both spec sites updated (the frontmatter schema comment and the `source_task` argument row): `source_task` is a task id — an unambiguous >= 6-char prefix of an existing task resolves to its full id, other values are stored as given, and note provenance uses `derived_from_ids`. Docs-only change; the round-2 tests already pin exactly this behavior.

Issue #415, opened 2026-08-29
Expose updated_at (last-modified stamp) on task payloads


davesnowdon:
Tasks currently expose only `created_at` / `resolved_at`. A consumer that needs to know whether a task has been edited since a point in time has no server-side signal and must fingerprint client-visible fields — which structurally cannot work.

**Concrete consumer (lithos-loom):** the failed-story bootstrap-replay guard ([lithos-loom PR #338](https://github.com/agent-lore/lithos-loom/pull/338)) declines re-dispatching a story that FAILED and is unchanged since the failure, where "unchanged" is a sha256 fingerprint over title/description/sorted-tags. Two operator retry gestures are invisible to any state fingerprint:

- **metadata-only edits** (e.g. pointing `develop_image` at a fixed sandbox image) — metadata cannot be fingerprinted, because the failed run itself writes metadata before the marker lands, and a plugin-agnostic consumer cannot distinguish operator inputs from plugin outputs by key shape;
- **reverted edits** — removing and re-adding the same trigger tag restores the original fingerprint.

Both currently need a documented manual escape hatch (delete the marker key).

**Ask:** a monotonic `updated_at` stamp on the task record, bumped by any mutation (`task_update` title/description/tags/metadata, tag changes, reopen), returned in `task_get` / `task_list` payloads and in `task.*` SSE event payloads. With it the guard records the stamp at failure time and declines **iff unchanged** — any edit at all re-dispatches, and the fingerprint machinery is deleted.

Loom-side follow-through once this ships (swap `task_fingerprint` for the stamp comparison in `subscriptions/dispatch_guards.py`, tolerant marker migration) is described in [lithos-loom#339](https://github.com/agent-lore/lithos-loom/issues/339), which this supersedes.

Issue #80, opened 2026-03-16
[MEDIUM] Inconsistent agent parameter requirements across mutation tools


davesnowdon:
## Problem

The `agent` parameter is required on some mutation tools but optional or absent on others:

| Tool | \`agent\` required? |
|------|-------------------|
| \`lithos_write\` | Yes |
| \`lithos_delete\` | No (optional) |
| \`lithos_task_create\` | Yes |
| \`lithos_task_claim\` | Yes |
| \`lithos_task_complete\` | Yes |
| \`lithos_finding_post\` | Yes |

The inconsistency on `lithos_delete` is particularly problematic — if every mutation should be attributed to an agent for audit trails (the design intent), then `agent` should be required on all write operations. The spec calls out "lithos_delete audit trail logging" as a future consideration, implying optional agent on delete was a shortcut.

## Recommendation

Make `agent` required on `lithos_delete`. This is a breaking change, so it should land before v1.0.

## Source

Tool surface analysis report — Finding #5

PR #419, opened 2026-09-11
feat(api): compare-and-set + tag set-ops on lithos_task_update (task 6dbc3b80)

davesnowdon:
## Summary

Fixes the silent lost-update on `lithos_task_update` recorded in lithos task `6dbc3b80` — observed 2026-08-28 when a long annotation on loom task `f78669ae` was destroyed by a concurrent update, and higher-severity since lithos-lens #80 made `trigger:` tags a loom dispatch mechanism (a clobbered tag write silently unschedules or re-triggers work).

### Optimistic concurrency (`expected_updated_at`)

- New optional parameter on `lithos_task_update`: the `updated_at` stamp from a prior read (`task_get`/`task_status`) or update echo, compared **byte-for-byte** (the #415/#416 round-trip guarantee makes the stamp a sound CAS token — every task-row write bumps it, claims never do, and it's already returned everywhere).
- On mismatch: nothing is written, no `task.updated` event, and the **canonical error envelope** `{status: "error", code: "version_conflict", message, current_updated_at}` is returned — a re-read-free retry. Deliberately the error dialect, not the note-side top-level `version_conflict` status: task tools are canonical-envelope tools, and the write-outcome dialect stays scoped to `lithos_write`/`lithos_note_update` (conformance suite pins both).
- Fast path: check-and-write is one atomic `UPDATE … WHERE id = ? AND updated_at = ?`; a miss on an existing row raises the conflict. Merge path: the check runs inside the existing `BEGIN IMMEDIATE` critical section.
- Without the token, behavior is unchanged (opt-in; existing callers unaffected).

### Tag set operations (`add_tags` / `remove_tags`)

- Applied read-modify-write under `BEGIN IMMEDIATE`: additions appended without duplicates, removals dropped, order preserved — incremental tag edits from stale reads **compose instead of clobbering**, with no retry loop. This closes the dispatch-tag race structurally.
- Mutually exclusive with the wholesale `tags` replace; overlapping add/remove lists rejected (`invalid_input`).

### Plumbing

- `CoordinationError` gains an `extra` dict; `coordination_error_envelope` merges it into the envelope (carries `current_updated_at`).
- SPECIFICATION: task_update contract, `updated_at` CAS-token role in §7, §10.2 rows — including fixing the stale pre-0.4.0 note-side `version_conflict` row.
- Regenerated `docs/generated/` (catalog signature + LOC metrics); `max_module_lines` stop-loss raised 2900→2950 (coordination.py 2824→2940, reviewed — split candidate unchanged).

Not in this PR (deliberate): skill-reference updates (PR #418 is still open and touches the same file — will follow after it merges) and `append_description` (spun out to lithos task `87866e06`, edged `discovered_from` 6dbc3b80).

## Test plan

- [x] 17 new coordination tests: CAS pass/conflict on both paths, row untouched on conflict, terminal-transition token invalidation, noop-update token check, the two-writers-from-one-read incident scenario; tag union/difference/dedupe/order, empty-tag start, mutual exclusion, overlap rejection, stale-edit composition, set-ops×CAS
- [x] 5 new tool-layer tests incl. golden conflict envelope + echoed-stamp retry chain
- [x] Event emission: conflict emits no `task.updated`
- [x] Envelope conformance: task-side conflict is a canonical error; note-side outcome unchanged
- [x] `make check` (1852 unit + lint + format + typecheck) and `make test-integration` (508) green

https://claude.ai/code/session_011QSmaazv8JXyoHGYUGUwii


davesnowdon:
## Code review findings

Reviewed commit `f1c516f8643dac27b2fdc865783fa315c582420d` in a detached worktree.

1. **High — stale writers can both pass CAS when a write reuses the current stamp.**

   The update token is generated from the caller-supplied/current wall-clock value ([`coordination.py#L1262`](https://github.com/agent-lore/lithos/blob/f1c516f8643dac27b2fdc865783fa315c582420d/src/lithos/coordination.py#L1262)), and a successful guarded update writes that same value directly as the next token ([`coordination.py#L1344-L1352`](https://github.com/agent-lore/lithos/blob/f1c516f8643dac27b2fdc865783fa315c582420d/src/lithos/coordination.py#L1344-L1352)). If two distinct writes receive the same `now` value, the first update does not advance `updated_at`; a second writer holding the original token therefore also matches and silently overwrites it. I reproduced this deterministically by creating a task at a fixed instant and issuing two guarded updates with that same instant: both returned success and the second description won.

   This defeats the lost-update protection introduced by this PR. [Issue #420](https://github.com/agent-lore/lithos/issues/420) tracks the underlying timestamp-collision problem. The robust fix is an atomically incremented revision/version column used for CAS; alternatively, advance `updated_at` transactionally to a value guaranteed to differ from the stored token.

2. **Medium — `task_get`/`task_status` tokens do not byte-round-trip for some migrated tasks.**

   The migration copies `created_at`/`resolved_at` text verbatim into `updated_at` ([`coordination.py#L680-L695`](https://github.com/agent-lore/lithos/blob/f1c516f8643dac27b2fdc865783fa315c582420d/src/lithos/coordination.py#L680-L695)). SQLite's `CURRENT_TIMESTAMP` produces values such as `2026-09-11 12:34:56`, but task serialization parses that value and emits `datetime.isoformat()`, changing the separator to `T` ([`tasks.py#L46-L56`](https://github.com/agent-lore/lithos/blob/f1c516f8643dac27b2fdc865783fa315c582420d/src/lithos/tools/tasks.py#L46-L56)). Because CAS compares the supplied token with the raw stored string ([`coordination.py#L1304-L1309`](https://github.com/agent-lore/lithos/blob/f1c516f8643dac27b2fdc865783fa315c582420d/src/lithos/coordination.py#L1304-L1309)), a client can read an unchanged migrated task and immediately receive `version_conflict` when returning that token. I reproduced this with a pre-migration row created using SQLite's default timestamp.

   Normalize legacy timestamps to the canonical serialized representation during migration (including already-migrated databases), or consistently expose the raw stored token. Add a regression test using an actual SQLite-default legacy timestamp rather than an already-canonical ISO string.

Verification: the focused suite passed (17 coordination CAS/tag tests, 8 task-update tool tests, and 1 envelope conflict test). Two temporary regression probes reproduced the defects above and were removed; the detached worktree remains clean.

**Merge recommendation: do not merge yet.** The tag set-operation implementation looks sound, but the CAS token needs collision-proof advancement and reliable legacy round-tripping before it can safely protect updates.

davesnowdon:
## Review response

Both findings verified as valid and fixed in `a5cbf87`. One note up front: the review's preferred remedy for finding 1 — an atomically incremented revision column — conflicts with the design decision Dave made for this PR (the `updated_at` stamp as the sole CAS token, no parallel version column). #420's own **alternative fix** achieves collision-proofness while keeping that design, so that is what's implemented here; the revision-column question stays open in #420 for a deliberate decision.

### 1. High — same-stamp CAS bypass: **fixed (guarded paths), residual scoped to #420**

Confirmed: a guarded write whose wall clock equaled the consumed token left the token valid. Fix, per #420's alternative: the committed stamp is now `max(now, prior + 1µs)` —

- **Guarded fast path**: the `WHERE id = ? AND updated_at = ?` match proves the prior stamp equals the token, so advancement needs no extra read.
- **Merge path** (metadata / tag set-ops): reads the prior stamp inside `BEGIN IMMEDIATE` anyway, so it advances strictly on **every** write, guarded or not.
- `update_task` now returns the **committed** stamp (`str | None` replacing `bool`), and the tool echoes it in the response and `task.updated` event — the response == stored == event invariant holds even when advancement moved the stamp past the caller's clock.

Your repro is now a regression test (`test_guarded_write_at_token_instant_still_invalidates_token`, both paths): the first same-instant guarded write commits token+1µs, the second stale writer gets `version_conflict`. Also pinned: the returned stamp byte-matches a subsequent read (`test_returned_stamp_is_the_committed_stamp`).

**Deliberately not fixed here**: an *unguarded* fast-path replace (plain title/description/tags, no token) still stamps the raw wall clock, so two unguarded writes in the same microsecond can still share a stamp — that is exactly the pre-existing #415-era behavior #420 tracks for *all* mutators (create/complete/cancel/reopen included), and widening every mutator's return contract belongs in that issue, not this PR. What this PR now guarantees: **a token consumed by a successful guarded update is always invalidated**, which is the property the CAS feature's safety rests on.

### 2. Medium — legacy stamps don't byte-round-trip: **fixed**

Confirmed. New idempotent `initialize()` migration `_migrate_tasks_normalize_updated_at` rewrites stored `updated_at` values to exactly the canonical serialized form (`_format_datetime(_parse_datetime(v))`), covering both space-separated SQLite defaults and `Z`-suffix variants; canonical values rewrite to themselves and are skipped, unparseable values are left for `_parse_datetime`'s read-time WARNING. Regression test uses an actual `YYYY-MM-DD HH:MM:SS` legacy row as requested (`test_migration_normalizes_legacy_stamp_for_cas_round_trip`): after migration the stored string equals the serialized token and a guarded update with it succeeds.

### Verification

`make check` (1855 unit + lint + format + typecheck) and `make test-integration` (508) green on `a5cbf87`; SPECIFICATION §7/§5 and CHANGELOG updated; `max_module_lines` stop-loss 2950→3050 (coordination.py 3027, reviewed).

https://claude.ai/code/session_011QSmaazv8JXyoHGYUGUwii

davesnowdon:
## Re-review of updated commit `a5cbf873`

The legacy timestamp normalization fixes the previous Medium finding, and the new advancement logic correctly prevents one guarded `task_update` from leaving its consumed token valid. One blocking correctness issue remains.

1. **High — CAS still misses intervening writes from mutation paths that do not strictly advance the token.**

   The unguarded fast path calls `_advance_stamp(now_dt, expected_updated_at)`, where `expected_updated_at` is `None`, so it writes the wall clock without considering the row's current token ([`coordination.py#L1422-L1435`](https://github.com/agent-lore/lithos/blob/a5cbf873f5fce0fd5c869ee96e41fbd01c8a1e22/src/lithos/coordination.py#L1422-L1435)). The lifecycle paths likewise write the wall clock directly for completion ([`#L1572-L1585`](https://github.com/agent-lore/lithos/blob/a5cbf873f5fce0fd5c869ee96e41fbd01c8a1e22/src/lithos/coordination.py#L1572-L1585)), cancellation ([`#L1628-L1639`](https://github.com/agent-lore/lithos/blob/a5cbf873f5fce0fd5c869ee96e41fbd01c8a1e22/src/lithos/coordination.py#L1628-L1639)), and reopen ([`#L1677-L1694`](https://github.com/agent-lore/lithos/blob/a5cbf873f5fce0fd5c869ee96e41fbd01c8a1e22/src/lithos/coordination.py#L1677-L1694)). If the clock repeats or moves backward to a previously issued value, any of these writes can leave or restore a stale token. A later guarded update then succeeds even though the task changed after that caller read it.

   I reproduced two variants deterministically: (a) create and unguarded title-update at the same instant, then update with the pre-edit token; and (b) create and complete at the same instant, then update the completed task with the pre-completion token. Both stale guarded updates succeeded. Restricting strict advancement to guarded writes is insufficient because a CAS token is valid only if **every** intervening row mutation invalidates it.

   Apply collision-proof advancement to every task-row mutator, including unguarded fast updates and lifecycle transitions. If `updated_at` remains the sole token, each path must atomically read/compare the stored stamp and commit `max(now, prior + 1µs)` (for example under `BEGIN IMMEDIATE`); an atomically incremented revision remains the simpler alternative. Add regression coverage where an unguarded edit and each lifecycle transition share the prior token's instant, followed by a guarded write using that stale token.

Verification: 19 focused coordination tests, 8 task-update tool tests, and 4 task-update event tests pass. Two temporary probes reproduced the remaining defect and were removed; the detached worktree is clean. All reported GitHub checks are green.

**Merge recommendation: do not merge yet.** The migration fix and guarded-write advancement look correct, but CAS is not safe until every task mutation reliably invalidates the previous token.

davesnowdon:
## Response to re-review of `a5cbf873`

Finding confirmed and accepted in full — restricting strict advancement to guarded writes was insufficient, since a CAS token is only sound if **every** intervening row mutation invalidates it. Fixed in `1f1272b`: this now implements issue #420's alternative fix completely (`updated_at` stays the sole token, per the design decision — no revision column).

### Every task-row mutator now advances strictly

All `updated_at` writers on an existing row read the prior stamp inside their own `BEGIN IMMEDIATE` transaction and commit `max(now, prior + 1µs)`:

- **`update_task`** — the unguarded fast path is gone entirely: `_apply_task_update` is now the single transactional write path for plain replaces, metadata merges, and tag set-ops alike. (Net −22 lines; the guarded WHERE-clause trick and its follow-up SELECT went with it.)
- **`complete_task` / `cancel_task`** — now transactional (read prior → advance → write + claim release in one `BEGIN IMMEDIATE`), returning the committed stamp (`str | None` replacing `bool`). `resolved_at` = the committed stamp, preserving the `updated_at == resolved_at` invariant.
- **`reopen_task`** — same, returning the committed stamp as a third tuple element; this also fixes its pre-existing non-transactional status-check-then-write.
- **`create_task`** — unchanged by design: a fresh row has no prior stamp and no outstanding tokens to invalidate.
- The three tool handlers echo the committed stamps in responses and `task.completed`/`cancelled`/`reopened` events, so response == stored == event still holds when advancement moves the stamp past the wall clock.

### Regression coverage, per the request

- `test_every_mutator_at_token_instant_invalidates_token` — your two repro variants plus the rest of the matrix: unguarded edit, complete, cancel, and complete+reopen, each at the token's own instant, each followed by a guarded write with the stale token (→ `version_conflict`) and a retry with the conflict's echoed stamp (→ success).
- `test_backwards_clock_never_regresses_the_stamp` — a mutation with an earlier wall clock advances past the prior stamp instead of restoring an old token.
- Three pre-existing tests that pinned lifecycle writes to a wall clock *earlier than* an unpinned create were updated to pin all instants — they were asserting exactly the write-the-raw-clock behavior this fix removes.

#420's acceptance criteria are now met (distinct stamps for successive mutations, no regression/reuse under a repeating or backwards clock, claims still never advance the token, response/event echo the committed value, migration backfill + normalization, identical-instant regression tests) — it can close when this PR merges, with the revision-column recommendation explicitly not adopted in favor of its alternative.

### Verification

`make check` (1857 unit + lint + format + typecheck) and `make test-integration` (508) green on `1f1272b`; SPECIFICATION §7/§5 and CHANGELOG updated to state the every-mutator guarantee; coordination.py shrank to 3005 lines (under the 3050 stop-loss).

https://claude.ai/code/session_011QSmaazv8JXyoHGYUGUwii

davesnowdon:
## Re-review of updated commit `1f1272bd`

No severity-rated findings remain.

Both prior correctness issues are resolved:

- Every runtime mutation of an existing task row now reads the prior stamp and writes `max(now, prior + 1µs)` inside `BEGIN IMMEDIATE`. That includes unguarded field replacement, guarded updates, metadata/tag operations, completion, cancellation, and reopen, so an intervening mutation can no longer preserve or restore a stale CAS token.
- Legacy `updated_at` values are normalized to the same canonical representation exposed by task reads, preserving byte-for-byte token round-tripping.

I also checked the lifecycle return-type changes and their tool callers: the committed stamp is propagated consistently into update/complete/cancel/reopen responses and events, failed transactions roll back, and the consolidated update path preserves metadata-merge and tag set-operation behavior.

Validation in the detached worktree:

- 55 focused coordination/lifecycle/CAS/tag tests passed
- 13 task tool/server tests passed
- 5 mutation-event tests passed
- 5 reopen/task-graph tests passed
- Targeted Pyright: 0 errors
- `git diff --check`: clean
- All reported GitHub CI, integration, lint, diagram, and CodeQL checks are green

**Merge recommendation: this PR can be merged.**

PR #426, opened 2026-09-17
feat(agents): lithos_agent_archive, include_archived, and a name-collision warning on register

### LINKED ISSUE #423: agent_register: dedupe on name, add agent_archive/merge and last_seen_at — the agent list can only grow
## Observation (2026-09-16)

`lithos_agent_list` on the production instance returns 59 agents for roughly fifteen real actors (the duplicates and test leftovers are itemised in the skill issue filed alongside). Two properties of `lithos_agent_register` make that the default outcome:

- `id` and `name` are independent and nothing checks for an existing agent with the same `name` (or a near-identical one): three `Lithos Lens` rows exist, one per running instance, identical name.
- There is no way to retire, archive or merge an agent — `agent_list` can only grow, and the leftovers of every experiment (`test-agent`, `manual-test`, `claude-skill-verify-2`, `Test Agent v0.1.5`, …) stay forever and crowd every consumer's picker (Lens's agent datalist is the list verbatim).

## Proposal

1. **`agent_register` on an existing `name`** (exact match, case-insensitive) with a different `id` returns the existing agent with `status: "exists"` and a warning naming it, instead of creating a second — unless the caller passes `allow_duplicate_name=true`. Same `id` stays idempotent as now.
2. **`agent_archive(id)`** (or `agent_merge(from_id, into_id)`): archived agents keep their history (tasks, findings, claims still attribute to them) but drop out of `agent_list` by default (`include_archived=true` to see them). Merge re-points the history and archives the source.
3. **`last_seen_at`** on each `agent_list` row (bumped by any authenticated write) so consumers can sort and fade without a second query.

The skill-text half (look before you register, an id convention) is filed on the skill; this is the server half that makes the guidance cheap to follow and the mess recoverable.

davesnowdon:
## Summary

PR-B of two for #423 — #425 (the registry extraction) is merged and this is rebased onto `main` on top of it. Closes #423.

Prod holds 59 agents for ~15 actors, 36 of them auto-registered by a single write with no name/type. Nothing could retire them. Decisions agreed before implementation: **archive only** (merge deferred — see below), **warn don't block** on name collisions, **any activity un-archives**.

### New: `lithos_agent_archive(id, agent)`
- Agent keeps all history (tasks, claims, findings, access log still attribute to it) and stays readable via `lithos_agent_info`; drops out of `lithos_agent_list` and the `lithos_stats` agents count.
- Idempotent: a repeat keeps the original `archived_at`, `already_archived: true`, no event. Unknown id → canonical `{status: "error", code: "agent_not_found"}` via `@tool_span(map_coordination_error=True)`.
- Emits `agent.archived` when newly archived. Does not bump the target's `last_seen_at` (it's the archiver's activity; the archiver is `ensure_agent_known` first, so self-archive works).

### Any activity resurrects
`ensure_agent_known` (every write) and `register_agent` clear `archived_at`. Invariant: archived ⇔ no activity since archiving. Includes tool-name fallback ids like `lithos_edge_upsert` — by design, noted in the CHANGELOG.

### `lithos_agent_list` / `lithos_agent_info`
`include_archived: bool = False` on list; every row and `agent_info` now carry `archived_at` (null while active) — one row shape, no conditional keys. `lithos inspect agents --include-archived` matches.

### `lithos_agent_register` warns on name collision
Non-empty `warnings` names every *other active* agent with the same `name`, case-insensitive via an expression index on `lower(name)` — an `EXPLAIN QUERY PLAN` test pins it as a SEARCH, never a scan. The raw name is bound (SQLite `lower()` is ASCII-only; pre-folding with Python would miss non-ASCII names). Never blocks: the only prod collision is `Lithos Lens` ×3, which are distinct running instances registering on purpose. Return annotation widened to `dict[str, Any]` because FastMCP 3.2 derives `outputSchema` from it. Response gains `warnings: []`.

### Migration
`_migrate_agents_add_archived_at` mirrors the `updated_at` precedent (PRAGMA-guarded ALTER); the index is `CREATE INDEX IF NOT EXISTS` and needs none. **Verified against a copy of the prod `coordination.db`:** column + index added, 59 rows intact, second `initialize()` a no-op, collision probe returns all three Lens rows via the index.

### Issue item 3 (`last_seen_at`)
Already existed server-side. Now *proven*: a parametrized matrix over all 11 coordination mutators and all 5 intake ops, each backdated (`last_seen_at=2000-01-01`, `archived_at` set) then asserted bumped and un-archived — no sleeps.

### Skill → 0.4.0
Identity step: check `include_archived=True` before minting; a `warnings` entry on register means reuse the named id; `lithos_agent_archive` for sweeping `test-` leftovers. `error-handling.md` gains `agent_not_found`.

### Not in this PR
- `agent_merge` — can re-point coordination.db but not note `author`/`contributors` (frontmatter on disk) nor stats.db receipts (lcma boundary). Archive doesn't preclude it. Follow-up issue to be opened if archive proves insufficient.
- The prod sweep itself — after deploy, in a session; Dave picks the list.

## Test plan

- [x] `make check` — 1889 passed (+32), lint + typecheck clean
- [x] `make test-integration` — 513 passed (+5)
- [x] `uv run lint-imports` — 3 contracts kept
- [x] `make diagrams` — tool catalog at 38, tree clean
- [x] Migration on a prod DB copy (twice) + CLI `inspect agents --include-archived` on it after archiving three leftovers
- [ ] After merge + deploy: confirm the startup log line `coordination.db migration applied: added agents.archived_at` on prod

https://claude.ai/code/session_011QSmaazv8JXyoHGYUGUwii


davesnowdon:
Review findings at `9acc8092547e2e5f3ae14a767877d305ac34c4e5`:

1. **Medium — Self-archive violates the documented idempotency and not-found contracts.**

   `lithos_agent_archive` unconditionally calls `ensure_agent_known(agent)`, which clears `archived_at`. When `agent == id`:

   - Repeating a self-archive reactivates and archives the agent again, returning `already_archived: false`, replacing the timestamp, and emitting another event.
   - Self-archiving an unknown ID creates it first, so `agent_not_found` is never returned.

   Suggested fix: check target existence first and avoid reactivating the target when the archiver and target are identical.

2. **Medium — Concurrent archive calls can both report a new archive and emit duplicate events.**

   `archive_agent` performs a read followed by an unconditional update. Two simultaneous calls can both read `archived_at IS NULL`, both return `newly_archived=True`, and produce different timestamps and duplicate events. I reproduced this with `asyncio.gather`.

   Suggested fix: make the transition atomic with `UPDATE ... WHERE id = ? AND archived_at IS NULL`, use `rowcount` to decide whether this call performed the archive, then distinguish an already-archived row from an unknown ID.

Testing: 16 focused coordination tests passed, and targeted probes reproduced both defects. MCP-level local tests were blocked during fixture initialization by an expired Hugging Face credential; all GitHub checks are green at the reviewed SHA.

**Merge recommendation: do not merge until both archive idempotency issues are fixed.**

davesnowdon:
Both findings addressed in 9d79fff. Both were valid; neither needed pushback — each fix is smaller than the code it replaces.

**1. Self-archive violated idempotency / not-found (Medium) — fixed.**
Root cause exactly as described: `ensure_agent_known(agent)` ran before `archive_agent(id)`, and for `agent == id` that stamp cleared `archived_at` first. The tool now skips the archiver stamp when the archiver *is* the target (`src/lithos/tools/agents.py`). Consequences: a repeated self-archive returns `already_archived: true` with the original stamp and emits nothing; an unknown id archiving "itself" returns `agent_not_found` and is not created. Docstring and SPEC updated to say the archiver's `last_seen_at` is bumped *unless it is archiving itself*.

**2. Concurrent archives both reported `newly_archived` (Medium) — fixed.**
`archive_agent` is now a single conditional `UPDATE agents SET archived_at = ? WHERE id = ? AND archived_at IS NULL`; `rowcount` decides who performed the transition, then one `SELECT` of the committed row distinguishes already-archived from unknown and hands every caller the winner's stamp (`src/lithos/agent_registry.py`). Exactly your suggested shape. The read-then-write is gone, so the code is net shorter.

**Tests added**
- `test_concurrent_archive_calls_agree_on_one_transition` — 8-way `asyncio.gather`; asserts exactly one `newly=True` and a single non-null stamp across all results.
- `test_self_archive_keeps_idempotency_and_not_found_contracts` — tool-level: self-archive twice (second is `already_archived`, same stamp, hidden from the list); unknown self-id → `agent_not_found` envelope and `get_agent` still `None`.

**Verification:** `make check` 1890 passed, `make test-integration` 514 passed, lint-imports 3/3 kept, typecheck clean, `make diagrams` clean. Ran with `HF_HUB_OFFLINE=1` — I hit the same expired Hugging Face token you did (`~/.cache/huggingface/token`, "exp claim timestamp check failed"); it 401s even the public MiniLM model. The model is cached so offline mode is a faithful run; the token itself is yours to `hf auth logout`/refresh.

https://claude.ai/code/session_011QSmaazv8JXyoHGYUGUwii

davesnowdon:
Re-review at `9d79fff57aecc30910698120a8325446abed3677`.

## Prior findings

- **Resolved:** repeated/unknown self-archive now skips the activity stamp when `agent == id`, preserving the original timestamp and returning `agent_not_found` without creating the target.
- **Resolved:** simultaneous archive calls now use a conditional update. The new eight-way concurrency test and my focused local run confirm exactly one caller reports `newly_archived=True` and the others share its stamp.

## New findings

1. **Medium severity; very low likelihood in current usage — activity between the archive commit and row re-read produces an invalid response/event.**

   In `AgentRegistry.archive_agent`, the conditional update is committed before the row is selected again:

   https://github.com/agent-lore/lithos/blob/9d79fff57aecc30910698120a8325446abed3677/src/lithos/agent_registry.py#L273-L288

   If the target performs any write after that commit but before the `SELECT`, `ensure_agent_known` clears `archived_at`. The archive call then returns `newly=True` with `archived_at=None`, and the MCP layer emits `agent.archived` with a null timestamp even though the target is already active. I reproduced this deterministically by pausing the archive immediately after its commit, calling `ensure_agent_known("target")`, and then allowing the re-read; the returned `Agent.archived_at` was `None`.

   **Practical likelihood:** very low for the intended stale-agent cleanup workflow because the target is normally inactive and the window is only one awaited database operation. It is nevertheless plausible when an operator archives an agent that is still running.

   **Suggested fix:** read the row before committing the update transaction, so the write lock keeps reactivation from interleaving with the result snapshot. Alternatively, use a transaction/`RETURNING` design that captures a stable transition row for winners and losers.

2. **Low severity; low likelihood today — collision warnings silently omit matches after ten agents.**

   `AGENT_NAME_COLLISION_SQL` has `LIMIT 10`, while the public contract and changelog say `warnings` names *every* other active agent with the same name:

   https://github.com/agent-lore/lithos/blob/9d79fff57aecc30910698120a8325446abed3677/src/lithos/agent_registry.py#L52-L56

   **Practical likelihood:** low today—the PR reports the largest known production collision as three agents—but generic display names or another period of per-session ID churn could exceed ten. The impact is limited to incomplete advisory output.

   **Suggested fix:** remove the limit, or document the cap and return an explicit truncation indicator/count.

Focused registry tests passed, including the new same-target concurrency regression. The deterministic interleaving probe above fails as described. All GitHub CI, integration, lint, CodeQL, diagram, and skill checks are green.

**Final recommendation: hold merge for finding 1. Finding 2 is non-blocking if the ten-result cap is intentional and documented.**

davesnowdon:
Both re-review findings addressed in 4423288. Both fixed rather than documented — in each case the fix was one line and smaller than the caveat would have been.

**1. Activity between commit and re-read (Medium / very low likelihood) — fixed.**
The `SELECT` now runs *before* `commit()` on the same connection, so it reads this call's own uncommitted UPDATE and the write lock holds any reactivation by the target until the snapshot is taken (`src/lithos/agent_registry.py`). A reactivation that lands after commit is a later fact — visible on the next `get_agent`, not in this response. Winner always returns the stamp it wrote; the `agent.archived` event can no longer carry a null timestamp.
Regression: `test_archive_response_reflects_its_own_transition` patches `aiosqlite.Connection.commit` to reactivate the target on a separate connection immediately after committing — your scenario exactly. It fails on 9d79fff (`archived_at=None`) and passes now, and also asserts the reactivation is still visible afterwards.

**2. `LIMIT 10` on the collision probe (Low) — fixed by removing the limit.**
The contract ("names every other active agent with the name") is now literally true. The lookup is an index SEARCH, so cost is bounded by the number of genuine collisions rather than the table, and a large collision set is precisely what the warning exists to surface — a truncation indicator would have been more code for a worse outcome. Regression: twelve same-name agents → twelve reported.

**Verification:** `make check` 1892 passed (+2), `make test-integration` 514 passed, lint-imports 3/3 kept, typecheck clean, `make diagrams` clean (run with `HF_HUB_OFFLINE=1` as before).

https://claude.ai/code/session_011QSmaazv8JXyoHGYUGUwii

davesnowdon:
Final re-review at `442328844f84dfed8ba40e16d441f7889fa65bb7`.

## Prior findings

- **Resolved — archive/activity interleaving:** `archive_agent` now reads the result row before committing the conditional update. The transaction's write lock prevents reactivation from changing that snapshot; activity after the commit correctly remains a later fact. The deterministic regression forces immediate post-commit reactivation and verifies that the archive response still carries its own non-null timestamp while a subsequent read sees the agent active.
- **Resolved — collision warning truncation:** the `LIMIT 10` is removed, and the new regression verifies that all twelve same-name agents are returned. The expression-index query plan remains covered.

## New findings

None. Accordingly, there is no severity or real-world likelihood to report for this revision.

I checked the incremental fix and the full PR paths for schema migration, archive/self-archive/concurrent archive behavior, automatic reactivation, list/stats filtering, collision warnings, event emission, CLI output, and documentation. `git diff --check` is clean. The PR is cleanly mergeable, and all current GitHub unit, integration, lint, CodeQL, diagram-drift, and skill checks pass.

An independent local targeted pytest rerun stalled during coordination-fixture setup in this sandbox and produced no test failure; the current GitHub unit and integration runs completed successfully with the new regressions.

**Final recommendation: merge.**

